PR #2, opened 2016-09-05
Sphinx build fail on travis-ci

cdeil:
@bsipicz @astrofrog - I'm getting this Sphinx error from astropy-helpers for the new `healpix` package I set up today:
https://travis-ci.org/cdeil/healpix/jobs/157661206#L605

Help!?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ matrix:
         # Check for sphinx doc build warnings - we do this first because it
         # may run for a long time
         - python: 2.7
-          env: SETUP_CMD='build_sphinx -w'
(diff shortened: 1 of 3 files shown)


cdeil:
I've attached a commit here that should fix the docs build.

astrofrog:
@cdeil - build_sphinx is deprecated *but* should still work for now, so I think this indicates a bug in the deprecation layer. I don't have time to look into it right now, but we should open an issue on astropy-helpers.

cdeil:
I'm merging this. `build_docs` works on Python 2 and 3.

@astrofrog - Thanks for your help!

PR #6, opened 2016-10-17
Add healpix code from Astrometry.net, plus its dependencies

dstndstn:
This was a minimal-work import of the healpix.c and healpix-utils.c code, plus its dependencies in the Astrometry.net repository.  There will be a fair bit of extra stuff dragged in that could be trimmed with a bit more work.  Delaying that work until a point where it is decided to use this code or not.


cdeil:
@dstndstn - Thank you very much!!!

I'm merging this now.

I'll probably have some time to get started on the Cython wrapper next weekend, but help is always welcome. I think the first step has to be to build via setup.py and to link against it.

cc @tboch

PR #8, opened 2017-09-10
Implement main functionality in Python package

astrofrog:
This is a work in progress:

* [x] Add a vectorized Cython interface for functions to convert from longitude/latitude to healpix nested/ring indices, with and without sub-pixel relative offsets
* [x] Add a function to find the resolution from n_side
* [x] Implement interpolation
* [x] Implement functions to convert indices between nested and ring conventions
* [x] Implement/add interface to functions to find neighbors
* [x] Use 64-bit int C functions
* [ ] Design medium and high-level API around Cython functions


astrofrog:
For clarity I think I'd like to merge this if tests pass and work on the Python interface more in a separate PR

PR #29, opened 2017-09-24
Fix remaining precision issues

### LINKED ISSUE #28: BUG: incorrect results for large values of nside
At the moment, the tests ``test_ang2pix`` and ``test_ang2pix`` are not run for values of ``nside`` greater than 2**22, because the results then start to diverge between this package and healpy. I think this is because of issues with 32- vs 64-bit ints in the C code. One solution might be to simply update the C code to use int64_t throughout rather than mix int and int64_t as we do currently. Lower priority is that ``test_nest2ring`` and ``test_ring2nest`` fail for ``nside=2**29``, probably for similar reasons to those above.

If anyone is interested in investigating, please feel free to take a look - you can reproduce the failures by setting the ``integers(...)`` calls in ``test_healpy.py`` to ``integers(0, 29)``.

astrofrog:
This fixes #28

### DIFF
--- a/cextern/astrometry.net/healpix.c
+++ b/cextern/astrometry.net/healpix.c
@@ -135,12 +135,12 @@ void healpixl_decompose_ring(int64_t hp, int Nside, int* p_ring, int* p_longind)
 	ns2 = Nside64 * Nside64;
 	if (hp < 2 * ns2) {
 		ring = (int)(0.5 + sqrt(0.25 + 0.5 * hp));
-		offset = 2 * ring * (ring - 1);
+		offset = 2 * (int64_t)ring * ((int64_t)ring - 1);
 		longind = hp - offset;
 	} else {
 		offset = 2 * Nside64 * (Nside64 - 1);
 		if (hp < 10 * ns2) {
-			ring = (int)((hp - offset) / (Nside * 4)) + Nside;
+			ring = (int)((hp - offset) / ((int64_t)Nside * 4) + (int64_t)Nside);
 			offset += 4 * (ring - Nside64) * Nside64;
 			longind = hp - offset;
 		} else {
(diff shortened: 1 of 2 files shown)


astrofrog:
There are still some real precision issues here in some cases. Here are the indices returned by ``pix2ang`` in the range:

```
lat_min = 89.999996
lat_max = 89.999997
lon_min = -5
lon_max = +5
```

**Original healpy**

healpy_true


**This package**

healpy_false

cdeil:
@astrofrog - There's fails on travis-ci, e.g.:
https://travis-ci.org/cdeil/healpix/jobs/279308716#L907

Are you running hypothesis tests in a deterministic way or non-deterministic?

We should run them in a deterministic way, having fails pop up from time to time in random places / times isn't good, no?

astrofrog:
Currently hypothesis is running in a non-deterministic way. I'm not sure how we make it run deterministically - is there a command-line option? If so, do you want to add it here?

cdeil:
@astrofrog - One way to make hypothesis run in a deterministic fashion is to add the `derandomize=True` option to all the calls to `@setttings` in `healpix/tests/test_healpy.py`:
```
@settings(max_examples=1000, derandomize=True)
```
I didn't apply it, because I wasn't 100% sure if you want it.

As mentioned in the hypothesis docs:
```
settings.define_setting(
    'derandomize',
    default=False,
    description="""
If this is True then hypothesis will run in deterministic mode
where each falsification uses a random number generator that is seeded
based on the hypothesis to falsify, which will be consistent across
multiple runs. This has the advantage that it will eliminate any
randomness from your tests, which may be preferable for some situations
. It does have the disadvantage of making your tests less likely to
find novel breakages.
"""
)
```
and for hypothesis to remain useful, one has to add `@example` for failing test cases one has found, as regression tests:
http://hypothesis.readthedocs.io/en/latest/details.html#providing-explicit-examples

My suggestion would to go go int64 only, and then to start adding `@example` if there's still any problematic test cases found.

astrofrog:
@cdeil - thanks for the clarifications! I've used ``derandomize`` here.

PR #30, opened 2017-09-25
Add docstrings to healpy interface functions

cdeil:
This PR adds minimal docstrings to the functions in `healpix.healpy`, linking to the equivalent function in `healpy`. I think this is very useful, for quick checking if the functions really are equivalent in signature or not. Plus they have good docstrings, so no need to duplicate those in our docs.

I also added a `Migrate` section to the `healpy` docs page, that mentions that one can do the one-line import change to `from healpix import healpy to hp`.

@astrofrog - Please review.

### DIFF
--- a/docs/healpy_compat.rst
+++ b/docs/healpy_compat.rst
@@ -7,6 +7,9 @@ a `healpy <http://healpy.readthedocs.io>`_-compatible interface in
 This is not the recommended interface, and is only provided as a convenience
 for packages that want to support both healpy and this package.
 
+Example
+-------
+
 As an example, the :func:`~healpix.healpy.pix2ang` function can be used to get
 the longitude/latitude of a given HEALPix pixel (by default using the 'ring'
 convention)::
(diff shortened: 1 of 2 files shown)


docs/healpy_compat.rst
```diff
@@ -20,3 +23,20 @@ which agrees exactly with the healpy function::
...
+To migrate a script or package from using ``healpy`` to this ``healpix`` package,
+to check if the required functionality is available by changing all::
+
+    import healpy as hp
+
+to::
+
+    from healpix import healpy as hp
+
+and see what's missing or brakes. Please file issues or feature requests!
```
astrofrog: typo: brakes -> breaks

healpix/healpy.py
```diff
@@ -16,11 +16,19 @@
...
+    'nside2npix',
+    'npix2nside',
+    'pix2ang',
+    'ang2pix',
+    'order2nside',
+]
 
 
 def nside2resol(nside, arcmin=False):
+    """Drop-in replacement for healpy `~healpy.pixelfunc.nside2resol`."""
```
astrofrog: Good idea! Does intersphinx work?

docs/healpy_compat.rst
cdeil: I'll amend the commit to fix the typo.

healpix/healpy.py
cdeil: Yes, I checked locally. Intersphinx to healpy works.

cdeil:
Should be OK to merge now.

PR #33, opened 2017-09-25
Improve order argument validation a bit

cdeil:
This PR improves argument validation for `order` a little bit, avoiding the many duplications of `ORDER[order.lower()]`.

I also changed the language from "should" to "must" in the errors, and would suggest to do this change for all the errors in this PR. @astrofrog - OK?

The main question here I think is whether we should call `.lower` and support `order='RING'` in addition to `order='ring'`? On the one hand I think being strict it better, because it means that there's one way to do it instead of a few and all scripts calling this will look the same. On the other hand, there might be a reason to support multiple capitalisations here? If there isn't a strong one, I'm +1 to be strict.

@astrofrog - Thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ distribute-*.tar.gz
 healpix/_compiler.c
 healpix/raw/wrap.c
 healpix/core_cython.c
+
+# pytest cache folder
+v
--- a/healpix/core.py
+++ b/healpix/core.py
@@ -22,9 +22,12 @@ __all__ = [
 
 
 def _order_str_to_int(order):
-    if order == 'nested':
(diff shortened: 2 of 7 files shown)


astrofrog:
@cdeil - the reason for supporting the uppercase one was because that's what is used in FITS files, and the examples become more complicated if we require users to do ``.lower()`` themselves. However, if you want we could support just ``ring`` and ``RING`` but not e.g. ``rINg``.

cdeil:
@astrofrog - How about now?

astrofrog:
@cdeil the Travis failure is due to the approximate keyword fix. The doctest result should be updated.

PR #35, opened 2017-09-25
Bug in cone_search_skycoord

cdeil:
@astrofrog - Here you're passing as forth argument `self.nside`, but I think it should be `approximate`, no?
https://github.com/cdeil/healpix/blob/a3cfdef81ee9d82c650ca9982cfd60897f2c12b7/healpix/high_level.py#L331

Would it be easy for you to fix and add a test case or should I do it?

### DIFF
--- a/healpix/high_level.py
+++ b/healpix/high_level.py
@@ -328,7 +328,7 @@ class CelestialHEALPix(HEALPix):
         skycoord = skycoord.transform_to(self.frame)
         representation = skycoord.represent_as(UnitSphericalRepresentation)
         lon, lat = representation.lon, representation.lat
-        return self.cone_search_lonlat(lon, lat, radius, self.nside)
+        return self.cone_search_lonlat(lon, lat, radius, approximate=approximate)
 
     def boundaries_skycoord(self, healpix_index, step):
         """


healpix/high_level.py
```diff
@@ -328,7 +328,7 @@ def cone_search_skycoord(self, skycoord, radius, approximate=False):
         skycoord = skycoord.transform_to(self.frame)
         representation = skycoord.represent_as(UnitSphericalRepresentation)
         lon, lat = representation.lon, representation.lat
-        return self.cone_search_lonlat(lon, lat, radius, self.nside)
+        return self.cone_search_lonlat(lon, lat, radius, approximate)
```
astrofrog: Could you use the keyword argument notation? (``approximate=approximate``). This will avoid future issues.

cdeil: Do you mean just for the last argument, or pass all arguments along as keyword arguments?
I usually pass everything as keyword arguments, but either way is fine with me.

astrofrog: I meant specifically ``approximate`` since it's an actual keyword argument in ``cone_search_lonlat``.

We could indeed expand to always specifying the argument name for all args, but then we should do it consistently. I think it's mostly important here since ``approximate`` is a kwarg and therefore API-wise we could choose to move it position-wise.

cdeil: OK, let's use the pattern to pass keyword arguments if the called function uses a keyword argument.

cdeil:
Merging, since there's no test for this case anyways.

astrofrog:
@cdeil - it is, indirectly ;) See failure here: https://travis-ci.org/cdeil/healpix/jobs/279531819

PR #42, opened 2017-09-28
Preserve shape of arrays in core Python functions

astrofrog:
@cdeil - I had a bit of time so worked on this. I have not done anything for the functions that get neighbours or boundaries since those already return 2D arrays, so I thought it might be confusing if we return e.g. an (8, 3, 2) array (where 8 is the number of neighbours). What do you think?

Fixes https://github.com/cdeil/healpix/issues/21

### DIFF
--- a/healpix/core.py
+++ b/healpix/core.py
@@ -21,6 +21,20 @@ __all__ = [
 ]
 
 
+def _restore_shape(*args, **kwargs):
+    shape = kwargs['shape']
+    if shape:
+        if len(args) > 1:
+            return [arg.reshape(shape) for arg in args]
+        else:
+            return args[0].reshape(shape)
+    else:
+        if len(args) > 1:
+            return [np.asscalar(arg) for arg in args]
+        else:
+            return np.asscalar(args[0])
+
+
 def _order_str_to_int(order):
     # We also support upper-case, to support directly the values
     # ORDERING = {'RING', 'NESTED'} in FITS headers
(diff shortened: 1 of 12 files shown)


astrofrog:
Btw the reason for importing six directly is because astropy.extern.six is going to be deprecated shortly.

healpix/core.py
```diff
@@ -341,7 +389,7 @@ def interpolate_bilinear_lonlat(lon, lat, values, order='ring'):
     if values.ndim != 1:
         raise ValueError("values must be a 1-dimensional array")
 
-    return core_cython.interpolate_bilinear_lonlat(lon, lat, values, order)
+    return _restore_shape(core_cython.interpolate_bilinear_lonlat(lon, lat, values, order), shape=shape)
```
cdeil: This function still contains the TODO and ndim check a few lines up.
It should be removed, no?

Matter of taste: I would separate the Cython call from the reshape, i.e. introduce an extra line here.

cdeil:
I forgot to say: the errors above are with this import:
```python
from healpix import healpy as hp
hp.ang2pix(nside=2, theta=0, phi=0)
hp.ang2pix(nside=2, theta=[[0]], phi=[[0]])
```
It works with healpy:
```python
>>> import healpy as hp
>>> hp.ang2pix(nside=2, theta=0, phi=0)
0
>>> hp.ang2pix(nside=2, theta=[[0]], phi=[[0]])
array([[0]])
```

astrofrog:
Ah oops the issue is that the healpy interface calls the cython functions directly. Let me fix that.

healpix/core.py
astrofrog: Sounds good, will fix

astrofrog:
@cdeil - the shape issues should be fixed and I also changed the Cython code here to use a string argument for order. Could you try the latest version to see if this works better for you now?

healpix/core.py
astrofrog: The TODO is about something else, it is if we want to support e.g. FITS files with vector columns for the actual map value.

cdeil:
@astrofrog - Thanks! The shape errors when running the hips tests are gone.

There's still this one missing function, so tests don't really run yet and I don't know if results are OK:
```
AttributeError: module 'healpix.healpy' has no attribute 'boundaries'
```
But that's unrelated. Suggest to merge this PR and continue in other PRs later.

cdeil:
I think you're missing the `oder = _validate_order(order)` line in some of the Cython functions, e.g. `healpix_with_offset_to_lonlat`?
That would mean that those functions don't work or give a good error with `order='ORDER'` in all-caps like the other functions, which is confusing to users.

astrofrog:
@cdeil - ok, I'll see if I can make a quick wrapper for that function (boundaries) - in a separate PR

astrofrog:
@cdeil - I'd rather not use the ``_validate`` function in Cython since it would mean importing a Python function into Cython (circular import) but I've just hard-coded the check with the uppercase version (which will be fastest).

cdeil:
Why not more the `_validate` function to Cython?

astrofrog:
There were some Python 2 issues - waiting for CI to pass then merging

PR #43, opened 2017-09-29
Added boundaries healpy compat function

astrofrog:
@cdeil - does this work for you?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,13 +78,11 @@ matrix:
 
         # Try older numpy versions
         - os: linux
+          env: PYTHON_VERSION=2.7 NUMPY_VERSION=1.12
+        - os: linux
+          env: PYTHON_VERSION=2.7 NUMPY_VERSION=1.11
+        - os: linux
           env: PYTHON_VERSION=2.7 NUMPY_VERSION=1.10
-        - os: linux
-          env: PYTHON_VERSION=2.7 NUMPY_VERSION=1.9
-        - os: linux
-          env: PYTHON_VERSION=2.7 NUMPY_VERSION=1.8
-        - os: linux
-          env: PYTHON_VERSION=2.7 NUMPY_VERSION=1.7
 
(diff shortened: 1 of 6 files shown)


cdeil:
It's now failing because `vec2ang` is missing in the healpy compat interface:
https://gist.github.com/cdeil/5cf1e3d66bbfa3ba270313034b3434a9

Since healpy returned `vec` for `boundaries`, having `boundaries` without that helper function isn't very useful. Can you add it?

And then, if tests pass, I would actually change the code in `hips` to not go via the `healpy` compat layer. But I think it's very valuable to have, because a one-line import change as a migration path to this new package is gold.

cdeil:
healpy boundaries returns a single Numpy array, and works for arrays of `pix`, not just scalar:
```
>>> from healpy import boundaries
>>> boundaries(2, [0, 1]).shape
(2, 3, 4)
```
http://healpy.readthedocs.io/en/latest/generated/healpy.boundaries.html

I think we would need the compat function to do the same, so that things really work for us in `hips`.

@astrofrog - OK to implement this here? Or should I stop using the healpy compat interface for `hips` and replace with the recommended interface now?

astrofrog:
Let's try and get this working here, this is a good test of the compat interface :) I'll try and get something working soon.

astrofrog:
@cdeil - can you try again?

astrofrog:
@cdeil - the failure is due to old versions of Numpy. Ok to drop Numpy < 1.10? (like for astropy core)

astrofrog:
I just saw hips requires Numpy > 1.11, so dropping Numpy < 1.10 here.

astrofrog:
Great! I'll just merge this now then 😄 

I would however recommend in future using the non-healpy API here, as you might find it does a lot of things for you, e.g. getting the skycoord for the corners of a pixel. This was a useful exercise though to build up the healpy compat interface.

PR #45, opened 2017-09-29
Remove approximate mode from cone search and clarify docstrings

astrofrog:
During investigations in https://github.com/cdeil/healpix/issues/41 I found that the approximate mode produces strange results, so I don't think we should expose it through the API. We could in future implement a mode equivalent to the one in healpy that uses just the center of the pixels, but that can be done without the C code.

### DIFF
--- a/healpix/core.py
+++ b/healpix/core.py
@@ -414,13 +414,14 @@ def healpix_neighbors(healpix_index, nside, order='ring'):
     return core_cython.healpix_neighbors(healpix_index, nside, order)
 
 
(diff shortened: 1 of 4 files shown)


cdeil:
👍 , thanks!

I hadn't understood what the "approximate mode" was before, and think it's better to start small and then add if necessary. I do think that there would be value in having `query_disc` in the healpy interface for the mode using the center of the pixels (and to have that option in this package), because it's frequently used and thus helps to migrate to this package. But it can be added later.

Issue #10, opened 2017-09-14
Rename to astropy-healpix


astrofrog:
If there is interest from the rest of the astropy community to move this to the astropy organization, I propose that we rename the repository and package to astropy-healpix.

cdeil:
@astrofrog - I've made a proposal in e364fe8 in  #47 how I would chose names / do the release:

![screen shot 2017-09-30 at 13 21 56](https://user-images.githubusercontent.com/852409/31045382-9bc8b290-a5e2-11e7-9e3a-e96410d8d29a.png)

It's the naming scheme I would prefer and also the minimal amount of work, given that the name of this repo, this package and the PyPI name and RTD name can stay as `healpix` as-is, and only a conda-forge package by the name of `astropy-healpix` needs to be added. It's also equivalent to what we did for `regions`.

That said, I don't care strongly about names, and @astrofrog - if you were willing to do renames to whatever you prefer, so that we can get a v0.1 release out today or at least soon, then please just go ahead.

astrofrog:
@cdeil - those kinds of inconsistencies bug me... For regions it's actually different since the RTD name is also different from the package (in fact I can't remember who I was speaking to but they said regions should have been called astropy-regions throughout).

This is also going to be an issue when releasing - we will then have to announce 'healpix' but then that might lead to confusion. I do think we should switch to astropy-healpix throughout, and I'm happy to do all the renaming.

One thing we could do to minimize pain is *also* switch this repo to the astropy organization today since then we avoid having to fix all the references twice. What do you think?

cdeil:
> regions should have been called astropy-regions throughout

Not possible, not a valid Python symbol one can import.
It's the same problem / confusing the `sklearn` / `scikit-learn` and other `scikit-` or `astropy-` packages have.
I'm sure you know, but just to be clear, doing the `scikit.learn` or `astropy.healpix` via trickery is the worst, let's not do that.

> One thing we could do to minimize pain is also switch this repo to the astropy organization today since then we avoid having to fix all the references twice. What do you think?

Please go ahead (and take over #47 and just put all the names (including RTD and repo) you prefer!

We should do all re-location and re-naming before the v0.1 release and then never change it again (until the point / if where we propose this package for inclusion in Astropy core).

astrofrog:
> Not possible, not a valid Python symbol one can import

Ah sorry I should have clarified - as with other packages that have a - in the name, the python module would have a _. Really scikit-learn should be scikit_learn for the import.

[...]

PR #47, opened 2017-09-30
Tidy up and rename package ahead of release

cdeil:
This PR improves the installation docs page a bit, adds a license line at the top of all files, and has some PEP8 whitespace formatting -> i.e. some cleanup towards a 0.1 release.


docs/installation.rst
```diff
@@ -40,8 +41,8 @@ Installing the latest stable version is possible either using pip or conda.
 Using pip
 ---------
 
-To install **healpix** with `pip <http://www.pip-installer.org/en/latest/>`__
-from `PyPI <https://pypi.python.org/pypi/regions>`__
+To install ``healpix`` with `pip <http://www.pip-installer.org/en/latest/>`__
```
astrofrog: In documentation I don't like using code notation to refer to the actual package (when not in the context of imports), hence why I use bold to distinguish it. This will be more apparent if we rename to ``astropy-healpix`` which looks like code but definitely isn't the import. What do you think?

healpix/tests/test_healpy.py
```diff
@@ -61,7 +61,8 @@ def test_npix2nside(npix):
 
 @given(nside_pow=integers(0, 29), nest=booleans(), lonlat=booleans(),
        lon=floats(0, 360, allow_nan=False, allow_infinity=False).filter(lambda lon: abs(lon) > 1e-10),
-       lat=floats(-90, 90, allow_nan=False, allow_infinity=False).filter(lambda lat: abs(lat) < 89.99 and abs(lat) > 1e-10))
+       lat=floats(-90, 90, allow_nan=False, allow_infinity=False).filter(
+           lambda lat: abs(lat) < 89.99 and abs(lat) > 1e-10))
```
astrofrog: This might look cleaner if you break before the ``.filter`` (which works)

astrofrog:
Once this passes, I'll merge and can then update the bit about the naming when I do the PR to rename.

astrofrog:
There are some failures due to a new example you added in the docs - so I'll go with your original suggestion and take over this PR for the renaming, and I can fix the issue at the same time.

astrofrog:
@cdeil - you can just add the changes relating to the API here if you don't want to open a separate PR

cdeil:
OK, I'll do a few more edits here now.

cdeil:
I've done a few changes here, moving towards what we discussed today in #10.

I have to go now. I might have time to do a little more coding tomorrow, but @astrofrog - if you could just merge my open PRs and go ahead with the move / rename, that would be great!

One thing I noticed is this doctest fail:
```
066 Testing installation
067 --------------------
068 
069 To check that you have this package installed and which version you're using,
070 start Python and execute the following code:
071 
072 .. code-block:: bash
073 
074     $ python
075     >>> import astropy_healpix
UNEXPECTED EXCEPTION: ModuleNotFoundError("No module named 'astropy_healpix'",)

  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/doctest.py", line 1330, in __run
    compileflags, 1), test.globs)

  File "<doctest installation.rst[0]>", line 1, in <module>

ModuleNotFoundError: No module named 'astropy_healpix'
```

I think this Astropy doctest plugin is a bit too relentless, should it really be running things in `code-block:: bash` by default?

astrofrog:
@cdeil - ah no I don't think it should be running on the bash code blocks... I'll try and finish this all up and do the renaming everywhere. Thanks!

PR #48, opened 2017-09-30
Do the six ourselves

cdeil:
While updating the install page in #47 I noticed that to be complete we should mention the dependency on six. But then I looked and all we need in `integer_types` in the tests. So I'd suggest to just avoid the dependency -> this PR.

@astrofrog - thoughts?


cdeil:
> can you remove six from the three CI config files?
done.

Issue #58, opened 2017-10-02
Use OpenMP in Cython wrappers


astrofrog:
Cython provides a way via ``prange`` to use OpenMP which we could use to speed up calculations since our loops are embarassingly parallel:

http://cython.readthedocs.io/en/latest/src/userguide/parallelism.html

The main trick will be to figure out how to determine whether the ``-fopenmp`` flag can be safely used in ``setup.py`` (this is not supported by all compilers)

(related to #22)

PR #59, opened 2017-10-02
Use OpenMP in Cython wrappers

astrofrog:
This is a WIP, but it seems to work:

**Without OpenMP compilation flags**

```
  fct    nest nside   size  time_healpy time_self   ratio 
------- ----- ----- ------- ----------- ---------- -------
pix2ang  True     1      10   0.0000133  0.0004341   32.70
pix2ang  True   128      10   0.0000100  0.0003775   37.76
pix2ang  True     1    1000   0.0000521  0.0005752   11.04
pix2ang  True   128    1000   0.0000386  0.0006398   16.56
pix2ang  True     1 1000000   0.0364820  0.1662130    4.56
pix2ang  True   128 1000000   0.0316353  0.1336160    4.22
pix2ang False     1      10   0.0000151  0.0005352   35.45
pix2ang False   128      10   0.0000124  0.0004764   38.47
pix2ang False     1    1000   0.0000462  0.0005917   12.82
pix2ang False   128    1000   0.0000592  0.0006090   10.29
pix2ang False     1 1000000   0.0385104  0.1494491    3.88
pix2ang False   128 1000000   0.0479088  0.1596320    3.33
```

**With OpenMP compilation flags**

```
  fct    nest nside   size  time_healpy time_self   ratio 
------- ----- ----- ------- ----------- ---------- -------
pix2ang  True     1      10   0.0000089  0.0009528  107.47
pix2ang  True   128      10   0.0000092  0.0007257   78.79
pix2ang  True     1    1000   0.0000426  0.0005569   13.07
pix2ang  True   128    1000   0.0000374  0.0005880   15.74
pix2ang  True     1 1000000   0.0588481  0.1137643    1.93
pix2ang  True   128 1000000   0.0431057  0.1032697    2.40
pix2ang False     1      10   0.0000103  0.0009352   90.46
pix2ang False   128      10   0.0000121  0.0006978   57.57
pix2ang False     1    1000   0.0000415  0.0007433   17.90
pix2ang False   128    1000   0.0000564  0.0007113   12.62
pix2ang False     1 1000000   0.0411127  0.0812799    1.98
pix2ang False   128 1000000   0.0498509  0.0795615    1.60
```

However there is still some work to do here. With this I get some occasional test failures suggesting things aren't quite right:

```
    def test_interpolate_bilinear_skycoord(self):
        values = np.ones(12 * 256 ** 2) * 3
        coord = SkyCoord([1, 2, 3] * u.deg, [4, 3, 1] * u.deg, frame='fk4')
        result = self.pix.interpolate_bilinear_skycoord(coord, values)
        assert_allclose(result, [3, 3, 3])
    
        # Make sure that coordinate system is correctly taken into account
    
        values = np.arange(12 * 256 ** 2) * 3
        coord = SkyCoord([1, 2, 3] * u.deg, [4, 3, 1] * u.deg, frame='fk4')
    
        result1 = self.pix.interpolate_bilinear_skycoord(coord, values)
        result2 = self.pix.interpolate_bilinear_skycoord(coord.icrs, values)
    
>       assert_allclose(result1, result2)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       (mismatch 33.33333333333333%)
E        x: array([ 1870655.960288,  1870303.535714,  1868938.874628])
E        y: array([ 1870683.430786,  1870303.535714,  1868938.874628])
```

(these are non-deterministic so maybe this has to do with variables being accessed/modified from different threads)

We also need to figure out how to determine if it's safe to put the -fopenmp flag or not (depends on the compiler). We might need a helper function that could live in astropy-helpers for that.

Closes https://github.com/astropy/astropy-healpix/issues/58

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "astropy_helpers"]
 	path = astropy_helpers
-	url = https://github.com/astropy/astropy-helpers.git
+	url = https://github.com/astrofrog/astropy-helpers.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "astropy_helpers"]
 	path = astropy_helpers
-	url = https://github.com/astrofrog/astropy-helpers.git
+	url = https://github.com/astropy/astropy-helpers.git
--- a/.run_docker_tests.sh
+++ b/.run_docker_tests.sh
@@ -16,6 +16,9 @@ uname -m
 echo "Output of sys.maxsize in Python:"
 python -c 'import sys; print(sys.maxsize)'
 
+apt-get update
+apt-get install -y git
+
 # We only test the docs output on 64-bit Linux so as not to have to 
 # degrade the output with ellipses to work on all platforms.
 python setup.py test -V -a "-s" --skip-docs
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 0.2 (unreleased)
 ================
 
-- No changes yet.
+- Use OpenMP to parallelize the Cython wrappers. [#59]
 
 0.1 (2017-10-01)
 ================
--- a/astropy_healpix/bench.py
+++ b/astropy_healpix/bench.py
@@ -88,7 +88,7 @@ def bench_run(fast=False):
                                           fast=fast)
 
                 results_single = dict(fct='pix2ang', size=int(size),
-                                      nside=nside, time_self=time_self)
+                                      nside=nside, time_self=time_self, nest=nest)
 
                 if HEALPY_INSTALLED:
                     time_healpy = bench_pix2ang(size=size, nside=nside,
--- a/astropy_healpix/conftest.py
+++ b/astropy_healpix/conftest.py
@@ -25,6 +25,7 @@ enable_deprecations_as_exceptions()
 try:
     PYTEST_HEADER_MODULES['Astropy'] = 'astropy'
     PYTEST_HEADER_MODULES['healpy'] = 'healpy'
+    PYTEST_HEADER_MODULES['Cython'] = 'cython'
     del PYTEST_HEADER_MODULES['h5py']
     del PYTEST_HEADER_MODULES['Pandas']
 except (NameError, KeyError):  # NameError is needed to support Astropy < 1.0
--- a/astropy_healpix/core_cython.pxd
+++ b/astropy_healpix/core_cython.pxd
@@ -8,16 +8,16 @@ from numpy cimport int64_t
 cdef extern from "healpix.h":
 
     # Converts a healpix index from the XY scheme to the RING scheme.
-    int64_t healpixl_xy_to_ring(int64_t hp, int Nside);
+    int64_t healpixl_xy_to_ring(int64_t hp, int Nside) nogil
 
     # Converts a healpix index from the RING scheme to the XY scheme.
-    int64_t healpixl_ring_to_xy(int64_t ring_index, int Nside);
+    int64_t healpixl_ring_to_xy(int64_t ring_index, int Nside) nogil
 
     # Converts a healpix index from the XY scheme to the NESTED scheme.
-    int64_t healpixl_xy_to_nested(int64_t hp, int Nside);
+    int64_t healpixl_xy_to_nested(int64_t hp, int Nside) nogil
 
     # Converts a healpix index from the NESTED scheme to the XY scheme.
-    int64_t healpixl_nested_to_xy(int64_t nested_index, int Nside);
+    int64_t healpixl_nested_to_xy(int64_t nested_index, int Nside) nogil
 
     #  Converts a healpix index, plus fractional offsets (dx,dy), into (x,y,z)
     #  coordinates on the unit sphere.  (dx,dy) must be in [0, 1].  (0.5, 0.5)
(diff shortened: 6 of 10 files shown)


astrofrog:
I want to get the OpenMP setup_pacakge.py helpers in astropy-helpers before proceeding here: https://github.com/astropy/astropy-helpers/pull/346

astrofrog:
@cdeil - I'll leave this open for a day or so in case you have any comments, but otherwise this can be merged.

cdeil:
@astrofrog - Thanks for implementing this!

Locally for me, at least with the default Apple clang compiler, I don't have OpenMP:
https://www.reddit.com/r/MacOS/comments/71hz1n/does_apple_bundled_clang_have_openmp_support_in/

But I see that it ran and works with openmp at least in this CI build:
https://travis-ci.org/astropy/astropy-healpix/jobs/286155579#L882

(Maybe `return 0` from the `test_openmp.c` to get rid of the compiler warning?)

So the summary of your timings above is that it's slower for small arrays, the same for mid-size ~1e3  arrays, and almost factor 2 faster for large ~1e6 size arrays. Is that with two cores or more?

---

Concerning this code changes in `astropy_healpix/core_cython.pyx`: I see that you define pointers `cdef double *dx_buf` at the top, and then further down in a `parallel` with block do
```
dx_buf = <double *> malloc(sizeof(double))
```
and an `if null abort` and later `free`.

Is this manual memory management really needed? Isn't there a way to allocate a temp array in a different way that is automatically memory-managed without explicit malloc / free calls? E.g. is it possible to make a temp array via Numpy?

If it has to be this way: did you ever test the "abort" case? What happens there? A Python MemoryError or an interpreter crash? Can you add a test or at least leave a comment in the code saying what this does (or pointing to the Cython docs page that explains it)?

Last question: does the user have any control over which cores are used? Either way, might be worth mentioning in the docs that by default multi-core is used. (I know at work many people run code on the main servers and then sometimes accidentally hammer the 48 core machines and cause problems with software that by default uses all cores).

astrofrog:
@cdeil - thanks for the review! I'll try and address all these points. Just a quick note if you want to try it out - if you install gcc with MacPorts you can do e.g.

```
CC=gcc-mp-6 python setup.py install
```

and it will compile with OpenMP enabled

cdeil:
Might be worth mentioning in the docs that the `CC` env variable is the one to use.

I see that on Windows, openmp is used and works:
https://ci.appveyor.com/project/Astropy/astropy-healpix/build/1.0.26/job/nismrvxs1joh1jqs
There's a ton of compiler warnings on Windows complaining about different integer sizes in different places and precision loss. Did you already go to all 64-bit int or is this something that still should be done?

On the one Mac build we have, the default Apple clang compiler is uses which doesn't have openmp (even if it shows up as gcc, it's really clang): https://travis-ci.org/astropy/astropy-healpix/jobs/286155578#L226 I think it's fine as-is, no need to extend the test setup. I just wanted to mention it here so that we have a record of what is / isn't tested.

cdeil:
Why is scipy not installed for the Mac test?
https://travis-ci.org/astropy/astropy-healpix/jobs/286155578#L307

And why doesn't the test runner say at the end how many tests were skipped?
https://travis-ci.org/astropy/astropy-healpix/jobs/286155578#L323

Locally I get `283 passed in 61.95 seconds` with this branch:
https://gist.github.com/cdeil/e232d793f88f04676a80d52e59eb9224#file-gistfile1-txt-L316

cdeil:
One more thing: as far as I can see, the build and test log no-where shows which Cython version is used.
@astrofrog - Can you add it here?
https://travis-ci.org/astropy/astropy-healpix/jobs/286155578#L306

cdeil:
Ah, I see - we don't use scipy at all. I thought we use it for interpolation.
In that case, all is good!

astrofrog:
> Locally for me, at least with the default Apple clang compiler, I don't have OpenMP:

Yes, Apple clang doesn't include OpenMP, but when we build conda packages we'll be able to use gcc I think. You can also compile with GCC yourself as specified in https://github.com/astropy/astropy-healpix/pull/59#issuecomment-335722487

> (Maybe return 0 from the test_openmp.c to get rid of the compiler warning?)

Done - this required an update in the helpers.

> So the summary of your timings above is that it's slower for small arrays, the same for mid-size ~1e3 arrays, and almost factor 2 faster for large ~1e6 size arrays. Is that with two cores or more?

That was with two threads (just using a laptop). I'd love to see results on a machine with more cores! But yes for small arrays the performance is slightly worse but the times are tiny so I don't think it will matter.

> Is this manual memory management really needed? Isn't there a way to allocate a temp array in a different way that is automatically memory-managed without explicit malloc / free calls? E.g. is it possible to make a temp array via Numpy?

We don't want to be using Numpy arrays here since we actually do want to just pass a simple C buffer (and making Numpy arrays would have a larger overhead). What I've done here follows the Cython guidelines, so I've added a link to those.

> If it has to be this way: did you ever test the "abort" case? What happens there? A Python MemoryError or an interpreter crash? Can you add a test or at least leave a comment in the code saying what this does (or pointing to the Cython docs page that explains it)?

As mentioned above I added a link. It would be really hard to trigger a crash because we're only allocating buffers with 8 elements at most, which will likely always work.

> Last question: does the user have any control over which cores are used? Either way, might be worth mentioning in the docs that by default multi-core is used. (I know at work many people run code on the main servers and then sometimes accidentally hammer the 48 core machines and cause problems with software that by default uses all cores).

Yes, I've added a documentation page to this PR that mentions this now (basically just set ``OMP_NUM_THREADS``. Though I'd love to see a multi-threaded result on a 48-core machine!

> Might be worth mentioning in the docs that the CC env variable is the one to use.

Done!

> There's a ton of compiler warnings on Windows complaining about different integer sizes in different places and precision loss. Did you already go to all 64-bit int or is this something that still should be done?

Not fully done yet but this is something that could be done in a separate PR. I'm not sure anymore if we really need to convert *all* integers to 64-bit though, but that's a discussion for another issue.

> And why doesn't the test runner say at the end how many tests were skipped?
https://travis-ci.org/astropy/astropy-healpix/jobs/286155578#L323

I think no tests were skipped here? (it doesn't say if none were skipped)

> One more thing: as far as I can see, the build and test log no-where shows which Cython version is used.

Fixed!

astrofrog:
@cdeil - let me know if you have any objections based on the above, or comments on the new performance doc page - otherwise I'll go ahead and merge this.

astrofrog:
Ok, thanks! Regarding the number of tests, it's possible you need to clean your local repo in case there are old test files lying around?

PR #61, opened 2017-10-10
Add neighbors method to high-level class

### LINKED ISSUE #52: healpix_neighbors missing from high-level class
The high-level HEALPix class is missing the ``healpix_neighbors`` method.

This should be straightforward to fix: every method in ``high_level.py`` has a matching function in ``core.py``, so this is just to make sure ``healpix_neighbors`` also exists in ``high_level.py``.

### LINKED ISSUE #53: healpix_neighbors should accept >1D arrays
``healpix_neighbors`` in ``astropy_healpix.core`` should be able to take arrays with more than one dimension for the pixel index - this is done in other functions in ``astropy_healpix/core.py`` but I think this one was missed. This issue can be tackled by someone new to the package.

astrofrog:
Fixes #52 - also renames the function and method to simply 'neighbors' since the ``healpix_`` is not meaningful here (we don't have it for e.g interpolate or cone search)

Also fixes #53

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,9 @@
 
 - Use OpenMP to parallelize the Cython wrappers. [#59]
 
+- Renamed the ``healpix_neighbors`` function to ``neighbors`` and added
+  a wrapper to the high-level class. [#61]
+
 0.1 (2017-10-01)
 ================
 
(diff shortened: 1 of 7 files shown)


astrofrog:
Interestingly healpy uses neighbours - shall I switch back for consistency with them and with the astrometry.net C code? (I actually like ``neighbours`` more even though I usually go with US spelling)

cdeil:
I think I would stick with "neighbours" then. But really, no strong opinion: do what you think is best!

astrofrog:
🇬🇧

Issue #55, opened 2017-10-01
Implement function to get interpolation weights


astrofrog:
We should implement a function equivalent to ``get_interp_weights`` from healpy

cc @woodmd

astrofrog:
Done in #63

Issue #54, opened 2017-10-01
Fix interpolation when there are not four neighbours present


astrofrog:
Currently, when doing the interpolation, the ``interpolate_bilinear_lonlat`` function in Cython searches for the four closest neighbours as if it was on a regular grid. However, in some places in the HEALPix grid, a given point is not surrounded by four cells, but three. The following example shows pixel index 653824 in a HEALPix grid with nside=256 in ring order, and its neighbours:

![debug](https://user-images.githubusercontent.com/314716/31053864-216a5d98-a69e-11e7-9ace-a57c9296690d.png)

The cross indicates the point where I was trying to interpolate, and where I got NaN with the current function. The position of the cross is lon=0.0 lat=-41.730291566995895

We should figure out what the correct behavior should be in this case - healpy deals with this case as follows:

```python
In [4]: healpy.get_interp_weights(256, 0.0, -41.730291566995895, nest=False, lonlat=True)
Out[4]: 
(array([653824, 653825, 655871, 654848]),
 array([ 0.40037327,  0.        ,  0.29981337,  0.29981337]))
```

I guess this must be doing some kind of [barycentric interpolation](https://en.wikipedia.org/wiki/Barycentric_coordinate_system#Interpolation_on_a_triangular_unstructured_grid)?

Currently if one of the four pixels to use for interpolation is -1 we just return NaN, but I think in that case we need to add a special case for 3-point interpolation. I *think* that this should then cover all cases.

cc @woodmd @cdeil

astrofrog:
Just another example - here it looks more complicated but I think it's still the same thing (3 neighbours instead of 4 at the location of the cross).

![debug](https://user-images.githubusercontent.com/314716/31058964-c9afa214-a6f4-11e7-8a11-1dc339c3d64c.png)

astrofrog:
Just FYI, I'm working on this!

astrofrog:
Done in #63 - it turns out my algorithm was completely different from healpy, so now this is implemented properly the issues described here are no longer relevant.

Issue #39, opened 2017-09-26
Test fails for test_pix2ang and test_ring2nest


cdeil:
I see two test fails locally. I think they are deterministic (tried two times and got the exact same numbers / fails), and probably precision-related. @astrofrog - Thoughts?
```
_________________________________________________________________________ test_pix2ang __________________________________________________________________________

    @given(nside_pow=integers(0, 29), nest=booleans(), lonlat=booleans(),
>          frac=floats(0, 1, allow_nan=False, allow_infinity=False).filter(lambda x: x < 1))

healpix/tests/test_healpy.py:75: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/hypothesis/executors.py:58: in default_new_style_executor
    return function(data)
/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/hypothesis/core.py:115: in run
    return test(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

nside_pow = 29, frac = 0.1666666694606345, nest = False, lonlat = False

    @given(nside_pow=integers(0, 29), nest=booleans(), lonlat=booleans(),
           frac=floats(0, 1, allow_nan=False, allow_infinity=False).filter(lambda x: x < 1))
    @settings(max_examples=2000, derandomize=True)
    def test_pix2ang(nside_pow, frac, nest, lonlat):
        nside = 2 ** nside_pow
        ipix = int(frac * 12 * nside ** 2)
        theta1, phi1 = hp_compat.pix2ang(nside, ipix, nest=nest, lonlat=lonlat)
[...]

PR #63, opened 2017-10-14
Fix bilinear interpolation

astrofrog:
This fixes both #54 (dealing properly with all corner cases) and #55 (added a function to get the interpolation weights) and also allows the interpolated array to be N-dimensional.

Things that still need to be done:

* [x] Better document the new C code
* [x] Fix healpy compat tests (results are mostly fine but there are precision issues in places)
* [x] Add some tests of the Python core functions

Other than that this is almost ready to go!


astrofrog:
@cdeil - are you interested in reviewing this or shall I just merge?

The interpolation and interpolation weights are now checked against healpy - I discovered that the way I was interpolating was quite different to healpy (and had issues e.g. #54) but now the two agree. The reason for writing the interpolation in C instead of Cython is because this routine is closely coupled with a number of other C functions from astrometry.net, so it was much cleaner to do it this way (and the C code is very simple anyway).

astrofrog:
Btw there are apparently still a few rare corner cases where healpy and this package disagree for the weights, but I think these are literal corner cases where a point is on a boundary between two pixels. Investigating.

astrofrog:
Since this fixes precision issues in the C healpix code, it may well have solved https://github.com/astropy/astropy-healpix/issues/39 - I added the examples from there here as known examples and they are passing.

cdeil:
I only reviewed the diff for 5 min. Tests and docstrings seem OK, I don't have the expertise / time to review the algorithm.  👍  to merge.

Just one comment: in `healpixl_compose_ring` the expressions are sprinkled with a lot of `(int64_t)` to cast ints to 64 bit. I think it would be easier to just take `int64_t` as input and avoid any int typecasting inside the functions.

Thanks!

astrofrog:
> Just one comment: in healpixl_compose_ring the expressions are sprinkled with a lot of (int64_t) to cast ints to 64 bit. I think it would be easier to just take int64_t as input and avoid any int typecasting inside the functions.

I agree, I'm planning another PR to just convert all the ints to int64_t to make this cleaner

PR #65, opened 2017-10-15
WIP: Use int64_t consistently in C code

### LINKED ISSUE #64: Convert C code to use only int64_t internally
At the moment there is a lot of typecasting going on in the C code between int and int64_t. As suggested by @cdeil it would be a lot simpler if we just used int64_t throughout, although there may be a few remaining cases where this is really not needed.

astrofrog:
Fixes #64

Still a work in progress at this point!

### DIFF
--- a/astropy_healpix/core_cython.pxd
+++ b/astropy_healpix/core_cython.pxd
@@ -8,27 +8,27 @@ from numpy cimport int64_t
 cdef extern from "healpix.h":
 
     # Converts a healpix index from the XY scheme to the RING scheme.
-    int64_t healpixl_xy_to_ring(int64_t hp, int Nside) nogil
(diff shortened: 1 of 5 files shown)


astrofrog:
Something isn't working right, so I'll need to investigate further, but no time for now

cdeil:
I only see `int` type changes in the diff. Any idea how that can lead to this change in shape?
https://travis-ci.org/astropy/astropy-healpix/jobs/288279333#L987
Or is this an unrelated change from upstream (i.e. in Numpy or Astropy)?

astrofrog:
I'm not sure what the failure here is due to. I'm pretty sure it was due to my changes here, I just haven't had a chance to investigate.

cdeil:
@astrofrog - I'm closing this old PR.
After #92 by @dstndstn there's too many merge conflicts, this will have to be re-applied.

PR #31, opened 2017-10-15
Add resolution to order helper function

cdeil:
Healpy has an `nside2resol` function:
http://healpy.readthedocs.io/en/latest/generated/healpy.pixelfunc.nside2resol.html

Both in Gammapy and the hips package we needed the reverse, a function that gives the closest nside or order for a given resolution.
* In Gammapy @woodmd implemented a `get_nside_from_pixel_size` function: https://github.com/gammapy/gammapy/blob/ae9067dad064aef16eb72e41cee425bf182b7229/gammapy/maps/hpx.py#L154
* In hipspy we added a `hips_order_for_pixel_resolution` function which is similar:
https://github.com/hipspy/hips/blob/2476251bea200d7221269b44772cb31660751730/hips/utils/healpix.py#L134

@astrofrog - Does it exist here already? Would it be welcome?

@astrofrog @woodmd - Can you recommend a signature? (i.e. return `nside` or `resolution`, but also if default should be to round up or down, or maybe even force the user to specify a `round={'down', 'up', 'float'}` or something like this by default?

Once I know what we want, I can add this in the coming days (with good docstring and tests).

### DIFF
--- a/astropy_healpix/core.py
+++ b/astropy_healpix/core.py
@@ -7,7 +7,7 @@ import math
 import numpy as np
 
 from astropy import units as u
-from astropy.coordinates import Longitude, Latitude, Angle
+from astropy.coordinates import Longitude, Latitude
 
 from . import core_cython
 from .core_cython import _validate_order
@@ -154,13 +154,14 @@ def pixel_resolution_to_nside(resolution, round='nearest'):
 
     Examples
     --------
+    >>> from astropy import units as u
     >>> from astropy_healpix import pixel_resolution_to_nside
-    >>> pixel_resolution_to_nside('13 arcmin')
+    >>> pixel_resolution_to_nside(13 * u.arcmin)
     256
-    >>> pixel_resolution_to_nside('13 arcmin', round='up')
+    >>> pixel_resolution_to_nside(13 * u.arcmin, round='up')
     512
     """
-    resolution = Angle(resolution).radian
+    resolution = resolution.to(u.rad).value
     pixel_area = resolution * resolution
     npix = 4 * math.pi / pixel_area
     nside = np.sqrt(npix / 12)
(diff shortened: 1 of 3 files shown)


cdeil:
>  why not just invert the equation for nside to resolution and then round the resulting nside value?

Good idea.

I've attached a commit with a function and test here doing that. (well almost, I'm rounding the equivalent "level", to make sure "nside" comes out as power of two again)

@astrofrog - Please review.

astropy_healpix/core.py
```diff
@@ -116,12 +117,72 @@ def nside_to_pixel_resolution(nside):
...
+    # we first compute the corresponding ``level = log2(nside)`
+    # round the level and then go back to nside
+    level = np.log2(nside)
+
+    if round == 'up':
+        level = np.array(level, dtype=int) + 1
+    elif round == 'nearest':
+        level = np.array(level + 0.5, dtype=int)
+    elif round == 'down':
+        level = np.array(level, dtype=int)
```
astrofrog: Rather than doing e.g. +1, +0.5, etc., what about using ``np.round(...)``, ``np.floor(...)`` and ``np.ceil(...)``? (then just add ``.astype(int)``. The issue is that rounding up currently won't give the optimal result if level happens to be an integer since it will return a resolution one too high (but ``ceil`` would do the right thing)

astropy_healpix/core.py
```diff
@@ -116,12 +117,72 @@ def nside_to_pixel_resolution(nside):
...
+
+    Examples
+    --------
+    >>> from astropy_healpix import pixel_resolution_to_nside
+    >>> pixel_resolution_to_nside('13 arcmin')
+    256
+    >>> pixel_resolution_to_nside('13 arcmin', round='up')
+    512
+    """
+    resolution = Angle(resolution).radian
```
astrofrog: Currently you say above that ``resolution`` should be a Quantity, but in the example/tests you use a string and you parse it here into an ``Angle`` object. While this can be convenient, I think at least in the Examples we should encourage people to use Quantity objects - elsewhere in astropy-healpix so far we just assume the input is a Quantity and will fail with a string.

Just to be clear, I could be convinced to allow strings as inputs but we should then update the Parameters section to mention this as an option, and probably show both string and Quantity in the Examples (and we should maybe do the same for other functions).

Finally, I think you can probably add ``copy=False`` to the Angle call to avoid an unnecessary copy when a Quantity is passed in.

astropy_healpix/core.py
cdeil: Done.

astropy_healpix/core.py
cdeil: Agreed that we should be consistent in our API and fully correct in the docstrings.
I've changed to the pattern you're using everywhere else for now, to require quantities to be passed in.

Adding support for strings for angle inputs as a convenience can be discussed in the future, and if we decide to add it, it should be done everywhere or it'll be confusing to users.

cdeil:
I've added a changelog entry. Merging.

PR #71, opened 2018-04-16
Support arrays for nside

### LINKED ISSUE #66: Support arrays for nside
One of the issues trying to switch Gammapy over to using astropy-healpix is that we call `ang2pix` with an array of nside values:
https://gist.github.com/cdeil/754b76dc7f22511a5504fbbe74dccd62#file-gistfile1-txt-L989

Minimum test case that works with healpy
```
>>> import healpy as hp
>>> hp.ang2pix(nside=[2, 4], theta=[0, 0], phi=[0, 0])
array([0, 0])
```
but doesn't work with `astropy_healpix.healpy` (casting to numpy arrays is missing):
```
>>> from astropy_healpix import healpy as hp
>>> hp.ang2pix(nside=[2, 4], theta=[0, 0], phi=[0, 0])
  File "<stdin>", line 1, in <module>
  File "/Users/deil/code/astropy-healpix/astropy_healpix/healpy.py", line 105, in ang2pix
    return lonlat_to_healpix(lon, lat, nside, order='nested' if nest else 'ring')
  File "/Users/deil/code/astropy-healpix/astropy_healpix/core.py", line 341, in lonlat_to_healpix
    nside = int(nside)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'list'
```
and also support for array-valued `nside` is missing:
```
>>> hp.ang2pix(nside=np.array([2, 4]), theta=np.array([0, 0]), phi=np.array([0, 0]))
  File "<stdin>", line 1, in <module>
  File "/Users/deil/code/astropy-healpix/astropy_healpix/healpy.py", line 105, in ang2pix
    return lonlat_to_healpix(lon, lat, nside, order='nested' if nest else 'ring')
  File "/Users/deil/code/astropy-healpix/astropy_healpix/core.py", line 341, in lonlat_to_healpix
    nside = int(nside)
TypeError: only length-1 arrays can be converted to Python scalars
```
I didn't look yet how much work it would be to make this work for `ang2pix` or possibly also similar other functions, I'm just going through and see what's missing for Gammapy today.

lpsinger:
Fixes #66.

### DIFF
--- a/astropy_healpix/core.py
+++ b/astropy_healpix/core.py
@@ -267,6 +267,7 @@ def healpix_to_lonlat(healpix_index, nside, dx=None, dy=None, order='ring'):
         raise ValueError('Either both or neither dx and dy must be specified')
 
     healpix_index = np.asarray(healpix_index, dtype=np.int64)
+    nside = np.asarray(nside, dtype=np.int64)
 
     if dx is None and dy is not None:
         dx = 0.5
@@ -282,13 +283,15 @@ def healpix_to_lonlat(healpix_index, nside, dx=None, dy=None, order='ring'):
         dx = dx.ravel()
         dy = dy.ravel()
 
-    shape = healpix_index.shape
-    healpix_index = healpix_index.ravel()
-    nside = int(nside)
-
     _validate_healpix_index('healpix_index', healpix_index, nside)
     _validate_nside(nside)
     order = _validate_order(order)
+
+    healpix_index, nside = np.broadcast_arrays(healpix_index, nside)
+
+    shape = healpix_index.shape
+    healpix_index = healpix_index.ravel()
+    nside = nside.ravel()
 
     if dx is None:
         lon, lat = core_cython.healpix_to_lonlat(healpix_index, nside, order)
@@ -331,14 +334,14 @@ def lonlat_to_healpix(lon, lat, nside, return_offsets=False, order='ring'):
     lon = lon.to(u.rad).value
     lat = lat.to(u.rad).value
 
-    lon, lat = np.broadcast_arrays(lon, lat)
+    lon, lat, nside = np.broadcast_arrays(lon, lat, nside)
 
     shape = np.shape(lon)
 
     lon = lon.astype(float).ravel()
     lat = lat.astype(float).ravel()
+    nside = nside.astype(np.int64).ravel()
 
-    nside = int(nside)
     _validate_nside(nside)
     order = _validate_order(order)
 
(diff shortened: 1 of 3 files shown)


cdeil:
@lpsinger - Thanks!
Changes look good to me.
Adding one very simple concrete test where you pass nside with arrays of different shapes would be useful IMO. But I think the tests where you `np.tile` nside actually do that, and if so, then OK to leave as-is.

I think these crashes are unrelated; they look like the Cython bug we had earlier this year:
https://travis-ci.org/astropy/astropy-healpix/jobs/367312532#L355
https://travis-ci.org/astropy/astropy-healpix/jobs/367312533#L1047

@bsipocz or @astrofrog - What do we have to do to resolve that? Update astropy-helpers?

lpsinger:
Side note: it would eliminate the need for a lot of the manual array broadcasting and datatype checks if the cython functions were implemented as ufuncs.

cdeil:
@lpsinger - Can you please make that suggestion about ufuncs in a new issue?
Otherwise there's the risk that this point just gets lost once this is merged.

Probably @astrofrog knows what you mean already, but for me or others, could you please include a link to a Cython example what exactly you propose to change to?

lpsinger:
> Probably @astrofrog knows what you mean already, but for me or others, could you please include a link to a Cython example what exactly you propose to change to?

Done. See #72.

cdeil:
The unrelated Cython crashes mentioned above have been fixed via #74 .
Restarting CI builds now.

cdeil:
Issues on Windows: https://ci.appveyor.com/project/Astropy/astropy-healpix/build/1.0.65/job/fmf107i99skhv740

@lpsinger - Can you have a look?

astrofrog:
I think it's the usual issue that Windows defaults to 32-bit ints

lpsinger:
Yuck. The only way around this that I see is to cast `nside` to `np.int64` in the tests.

I think that if we rewrote these functions as ufuncs (see #72) we would get casting like that automatically. @astrofrog, is this approach OK with you?

cdeil:
> The only way around this that I see is to cast nside to np.int64 in the tests.

This is just for tests that call directly into the Cython functions, and there you just have to pass the correct types, no? Users won't see these errors because they go through the Python wrapper functions which do the type casting to `np.int64` for them.

@lpsinger - Maybe do this change and merge this in? The discussion whether to use ufuncs instead and re-write the whole package will probably take longer (#72), and having the feature & tests here merged is more helpful than not, no?

One more thought: I see many `<int>` in your Cython code. Are those 64 bit? Related: #65 by @astrofrog .

lpsinger:
> @lpsinger - Maybe do this change and merge this in? The discussion whether to use ufuncs instead and re-write the whole package will probably take longer (#72), and having the feature & tests here merged is more helpful than not, no?

I would like to play around a bit longer with #72 first. I think that it may help with the int type too.

cdeil:
@lpsinger - We'll have the astropy-healpix coding sprint next Monday / Tuesday, and support for nside arrays is very high on the list of things we need. Do you have time to try and sort out the int here so that this can go in? Or do you really think the change to C / ufuncs should be done first before adding this?

lpsinger:
> @lpsinger - We'll have the astropy-healpix coding sprint next Monday / Tuesday, and support for nside arrays is very high on the list of things we need. Do you have time to try and sort out the int here so that this can go in?

Next week I will.

> Or do you really think the change to C / ufuncs should be done first before adding this?

I think that we should close pull request and just switch to ufuncs because this will kill two birds with one stone.

PR #73, opened 2018-04-17
Implement pix2vec, vec2pix, ang2vec

lpsinger:


cdeil:
@lpsinger - Thanks!

Adding tests to check that the new functions work for array inputs in the same way that healpy does, and yield output arrays with the same shape or even dtype would be nice.

But @lpsinger - I'd also be happy to merge this in as-is if you just want to keep as-is.

@astrofrog - Do you have time to review?

lpsinger:
> I agree with @cdeil that it would be nice to add a test or two to check that the shape is preserved

Done. Good thing too, because the shape was wrong before.

lpsinger:
FYI, the tests pass now.

PR #77, opened 2018-04-28
WIP: interpolate_bilinear_lonlat raises IndexError for nans

lpsinger:
The `interpolate_bilinear_lonlat` function raises an IndexError if the longitude or latitude that are passed to it are nans. This is because some or all of the indices returned by `bilinear_interpolation_weights` are negative in order to indicate failure, which may not necessarily be valid indices.

This breaks the HEALPix functions in the `reproject` package.

**Important note**: do not merge this pull request yet. This is a work in progress. It illustrates the bug by adding a unit test, but does not fix the bug.

### DIFF
--- a/astropy_healpix/tests/test_core.py
+++ b/astropy_healpix/tests/test_core.py
@@ -242,6 +242,13 @@ def test_interpolate_bilinear_invalid():
                                     values, order='banana')
     assert exc.value.args[0] == "order must be 'nested' or 'ring'"
 
+    result = interpolate_bilinear_lonlat([0, np.nan] * u.deg,
+                                         [0, np.nan] * u.deg, values,
+                                         order='nested')
+    assert result.shape == (2,)
+    assert result[0] == 1
+    assert np.isnan(result[1])
+
 
 def test_interpolate_bilinear_lonlat_shape():
 


cdeil:
@lpsinger or @astrofrog - How would you suggest to fix the issue?

I remember in Gammapy we also had issues with NaN when calling into scipy or other interpolation routines, and usually just did an array subset, interpolated on that, and then filled with 0 or NaN after.
(one example here: https://github.com/gammapy/gammapy/blob/8eb04cf4833667f878a4be5dd26137176f3ca6a8/gammapy/maps/wcsnd.py#L196)

But maybe in this case there's a better way by changing the lower-level functions?

astrofrog:
@cdeil - I think something like you suggest would be fine - for now a fix could go in the cython code, and in future in the ufunc code.

cdeil:
@lpsinger - Do you have time to finish this PR up, or should I try?
(let's try to finish up open PRs before the astropy-healpix coding sprint Monday / Tuesday)

lpsinger:
I'm sorry, I don't have time today. I can get to it by early next week. Please have a go at it.

lpsinger:
Should be good to go now.

cdeil:
Superseded by #106 .

Issue #81, opened 2018-05-18
Docstrings are too specific for 32-bit architectures


lpsinger:
Apparently, on 32-bit architectures, the string representation of Numpy arrays of type `int64` changes slightly.

See [Debian build log](https://buildd.debian.org/status/fetch.php?pkg=astropy-healpix&arch=i386&ver=0.2-2&stamp=1526595205&raw=0). This is for i386, but the error is similar for all supported 32-bit platforms. Excerpt here:

```
=================================== FAILURES ===================================
________________________ [doctest] docs/cone_search.rst ________________________
009 resolution HEALPix maps where the number of pixels may become arbitrarily large.
010 
011 Instead, the :meth:`~astropy_healpix.HEALPix.cone_search_lonlat` method can be used to
012 efficiently find all HEALpix pixels within a certain radius from a
013 longitude/latitude::
014 
015     >>> from astropy import units as u
016     >>> from astropy_healpix import HEALPix
017     >>> hp = HEALPix(nside=16, order='nested')
018     >>> hp.cone_search_lonlat(10 * u.deg, 30 * u.deg, radius=10 * u.deg)
Differences (unified diff with -expected +actual):
    @@ -1,3 +1,4 @@
     array([1269,  160,  162, 1271, 1270, 1268, 1246, 1247,  138,  139,  161,
            1245,  136,  137,  140,  142,  130,  131, 1239, 1244, 1238, 1241,
    -       1243, 1265, 1267, 1276, 1273, 1277,  168,  169,  163,  166,  164])
    +       1243, 1265, 1267, 1276, 1273, 1277,  168,  169,  163,  166,  164],
    +      dtype=int64)

/tmp/astropy_healpix-test-7Udyrf/lib/python2.7/site-packages/docs/cone_search.rst:18: DocTestFailure
________________________ [doctest] docs/coordinates.rst ________________________
029     >>> lon.hourangle  # doctest: +FLOAT_CMP
030     array([ 3.1875,  6.25  ,  1.8   ])
031     >>> lat.degree  # doctest: +FLOAT_CMP
032     array([  4.78019185,  54.3409123 , -44.99388015])
033 
034 Conversely, given longitudes and latitudes as :class:`~astropy.units.Quantity`
035 objects, it is possible to recover HEALPix pixel indices::
036 
037     >>> from astropy import units as u
038     >>> hp.lonlat_to_healpix([1, 3, 4] * u.deg, [5, 6, 9] * u.deg)
Expected:
[...]

PR #83, opened 2018-05-18
Updated CircleCI configuration to use better 32-bit docker image

astrofrog:
The 32-bit image we were using doesn't actually catch a lot of 32-bit issues (not sure why) so switching to the one we now use in astropy core.


astrofrog:
@lpsinger - "Good" news - the new config correctly picks up the failures you mentioned in https://github.com/astropy/astropy-healpix/issues/81 - I'm actually just going to merge it even though the CircleCI build fails as we can then fix it in a separate PR.

Issue #80, opened 2018-05-17
pix2ang fails on ppc64el


lpsinger:
See the [Debian build log]. Excerpt here:

```
=================================== FAILURES ===================================
_________________________________ test_pix2ang _________________________________

    @given(nside_pow=integers(0, 29), nest=booleans(), lonlat=booleans(),
>          frac=floats(0, 1, allow_nan=False, allow_infinity=False).filter(lambda x: x < 1))

astropy_healpix/tests/test_healpy.py:100: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3/dist-packages/hypothesis/core.py:579: in execute
    result = self.test_runner(data, run)
/usr/lib/python3/dist-packages/hypothesis/executors.py:58: in default_new_style_executor
    return function(data)
/usr/lib/python3/dist-packages/hypothesis/core.py:571: in run
    return test(*args, **kwargs)
astropy_healpix/tests/test_healpy.py:100: in test_pix2ang
    frac=floats(0, 1, allow_nan=False, allow_infinity=False).filter(lambda x: x < 1))
/usr/lib/python3/dist-packages/hypothesis/core.py:518: in test
    result = self.test(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

nside_pow = 27, frac = 0.6666666666666666, nest = True, lonlat = False

    @given(nside_pow=integers(0, 29), nest=booleans(), lonlat=booleans(),
           frac=floats(0, 1, allow_nan=False, allow_infinity=False).filter(lambda x: x < 1))
    @settings(max_examples=2000, derandomize=True)
    @example(nside_pow=29, frac=0.1666666694606345, nest=False, lonlat=False)
    def test_pix2ang(nside_pow, frac, nest, lonlat):
        nside = 2 ** nside_pow
        ipix = int(frac * 12 * nside ** 2)
        theta1, phi1 = hp_compat.pix2ang(nside, ipix, nest=nest, lonlat=lonlat)
        theta2, phi2 = hp.pix2ang(nside, ipix, nest=nest, lonlat=lonlat)
        if lonlat:
            assert_allclose(phi1, phi2, atol=1e-8)
[...]

Issue #38, opened 2017-09-25
Mention precision issues near the poles in docs


astrofrog:
Before we actually fix https://github.com/cdeil/healpix/issues/34 we should add a mention in the documentation about precision limitations - like healpy we are mostly limited to nside=29 (50mas resolution) but close to the poles (value tbd, but e.g. 89.999) the precision is worse here than in healpy.

Issue #34, opened 2017-09-25
Precision issues for large values of nside


astrofrog:
At the moment, our longitude/latitude to pixel conversion has a lower precision than that of the official HEALPix library. This is visible when making a map of pixel values for ``2**28``:

**Original HEALPix library (via healpy)**

![healpy_true](https://user-images.githubusercontent.com/314716/30810655-aeabe51e-a1fd-11e7-82d0-e956b59b1029.png)

**This package**

![healpy_false](https://user-images.githubusercontent.com/314716/30810664-b4bbafac-a1fd-11e7-8cd9-bad400e11f53.png)

I think I've tracked down the issue to the ``xyztohp`` function, and specifically these lines:

```
        // solve eqn 20: k = Ns - xx (in the northern hemi)
		root = (1.0 - vz*zfactor) * 3.0 * mysquare(Nside * (2.0 * phi_t - pi) / pi);
		kx = (root <= 0.0) ? 0.0 : sqrt(root);

        // solve eqn 19 for k = Ns - yy
		root = (1.0 - vz*zfactor) * 3.0 * mysquare(Nside * 2.0 * phi_t / pi);
		ky = (root <= 0.0) ? 0.0 : sqrt(root);

		if (north) {
			xx = Nside - kx;
			yy = Nside - ky;
		} else {
			xx = ky;
			yy = kx;
		}
```

and in particular the issue is with the ``x`` value (``y`` seems fine I think).

The issue is that doubles actually have less precision information than int64 (only 52 bits are allocated to significant precision), so by going through ``double`` and doing operations such as ``sqrt`` etc which further degrade precision, I **think** we are losing precision there. So this may not be a 32-bit vs 64-bit int issue but rather an issue with floating point precision. We should investigate whether we can improve the precision here, for example by using ``long double`` or finding other ways to avoid precision loss. For now, I will update the documentation to mention the resolution to which the results can be trusted.

@dstndstn - if you have any ideas on improving precision here, please feel free to comment!

dstndstn:
Hi,

If this is at Dec = 89.99999, then I'm guessing you're seeing cancellation
error in (1. - vz), where (I'm guessing) vz is the unit-sphere z vector --
which will be approaching 1 as Dec approaches 90 degrees.  Maybe you could
use a different trig identity here to avoid the cancellation error --
sin(Dec) = cos(pi/2 - Dec) -- for vz > 0.9 or 0.5 or whatever.

cheers,
--dustin



[...]

PR #84, opened 2018-05-18
Added failing example for test_pix2ang

astrofrog:
@lpsinger - https://github.com/astropy/astropy-healpix/issues/80 is not related to PPC, it is just a failure that happened to appear in that build. This PR adds that specific example to the hypothesis test and this fails for me locally on Mac. I don't have time to investigate why this is happening, so if anyone has any time please feel free to!

### DIFF
--- a/astropy_healpix/tests/test_healpy.py
+++ b/astropy_healpix/tests/test_healpy.py
@@ -100,6 +100,7 @@ def test_pix2ang_shape():
        frac=floats(0, 1, allow_nan=False, allow_infinity=False).filter(lambda x: x < 1))
 @settings(max_examples=2000, derandomize=True)
 @example(nside_pow=29, frac=0.1666666694606345, nest=False, lonlat=False)
+@example(nside_pow=27, frac=2./3., nest=True, lonlat=False)
 def test_pix2ang(nside_pow, frac, nest, lonlat):
     nside = 2 ** nside_pow
     ipix = int(frac * 12 * nside ** 2)


astrofrog:
This fails on Travis, as expected (I think healpy doesn't get installed on AppVeyor)

cdeil:
Here's a small example to reproduce the issue:
```
import numpy as np
import astropy_healpix

nside = 2 ** 27
frac = 2 / 3
ipix = int(frac * 12 * nside ** 2)
healpix = astropy_healpix.HEALPix(nside, order='nested')
lon, lat = healpix.healpix_to_lonlat(ipix)
phi = lon.rad
theta = np.pi / 2 - lat.rad

print('nside:', nside)
print('ipix=', ipix)
print('lat.dtype=', lat.dtype)
print('phi=', phi)
print('lat.deg=', lat.deg)
print('distance_to_south_pole=', np.pi - theta)
```
The problem is that this pixel is placed directly at the south pole:
```
nside: 134217728
ipix= 144115188075855872
lat.dtype= float64
phi= 0.0
lat.deg= -90.0
distance_to_south_pole= 0.0
```
healpy places is (presumably correctly) above the south pole
```
import healpy as hp
theta, phi = hp.pix2ang(nside, ipix, nest=True)
print(theta.dtype)
print('phi=', phi)
print('distance_to_south_pole=', np.pi - theta)
```
```
float64
phi= 0.7853981633974483
distance_to_south_pole= 6.083373360610267e-09
```

@dstndstn - could you  please try to run this example through only your C code? Is it a bug there, or is it a problem that's introduced in the Python wrapper? (e.g. because of `float` or `int` precision / size or rounding issues?)

cdeil:
@astrofrog - I now see #34 #38 . Looks like this case is just an example of a known issue: precision near the poles.

cdeil:
I restarted CI after merging #94.

Looks like the `phi` is still incorrect for this test case?
https://travis-ci.org/astropy/astropy-healpix/jobs/381277304#L1064

So @dstndstn @mreineck - looks like there are other precision issues in the C code.

dstndstn:
has anyone else had issues installing healpy on OSX?  After doing a "pip
install healpy --user", which seemed to succeed, I get

> python -c "import healpy"
/Users/dstn/Library/Python/3.6/lib/python/site-packages/healpy/pixelfunc.py:107:
UserWarning: Warning: cannot import _healpy_pixel_lib module
  warnings.warn("Warning: cannot import _healpy_pixel_lib module")
  File "<string>", line 1, in <module>
  File
"/Users/dstn/Library/Python/3.6/lib/python/site-packages/healpy/__init__.py",
line 81, in <module>
    from ._query_disc import query_disc, query_strip, query_polygon,
boundaries
ImportError:
dlopen(/Users/dstn/Library/Python/3.6/lib/python/site-packages/healpy/_
query_disc.cpython-36m-darwin.so, 2): Symbol not found:
__ZN14T_Healpix_BaseIlE8SetNsideEl23Healpix_Ordering_Scheme
  Referenced from:
/Users/dstn/Library/Python/3.6/lib/python/site-packages/healpy/_
query_disc.cpython-36m-darwin.so
  Expected in: flat namespace
 in /Users/dstn/Library/Python/3.6/lib/python/site-packages/healpy/_
query_disc.cpython-36m-darwin.so

astrofrog:
I often have issues with pip and have to resort to using the conda package from conda-forge.

dstndstn:
I looked into this failing test case.  It's very similar, in that there is cancellation error near the pole.

Specifically, the term being squared here is ~1e-9, so z -> 1 (cancellation error; it should be 1 - 1e-18).
https://github.com/astropy/astropy-healpix/blob/master/cextern/astrometry.net/healpix.c#L1013

dstndstn:
I've got a fix for this

cdeil:
@dstndstn - If you're willing to use conda, there's now a quick way to get an environment with `healpy` and everything you need for hacking on `astropy-healpix`: a0b72e7

My experience is that `pip install healpy` sometimes works and sometimes give linker errors like yours, but using conda and the healpy from conda-forge works reliably.

dstndstn:
Could you please check whether #101 fixes this?
I'm going to be away this weekend.

astrofrog:
I've rebased this, so it can be merged if CI passes

cdeil:
Looks like this case is working now. @dstndstn @astrofrog - Thanks!

PR #87, opened 2018-06-24
Fix make in  cextern/astrometry.net/Makefile

cdeil:
It would be nice to polish `cextern/astrometry.net` a bit as a C library (remove things that aren't needed, add a README, maybe move the `interpolate.c` here as well?

The first thing I wanted to do was build it and run the tests:
```
$ cd cextern/astrometry.net
$ make
cc -o qsort_reentrant.o -c qsort_reentrant.c
make: *** No rule to make target `cutest.o', needed by `test_healpix'.  Stop.
```
@dstndstn @astrofrog - Is that https://github.com/asimjalis/cutest ?
Should we use it here or is there something better to test a C lib?
If we use cutest: how? Should we bundle a copy of the files?https://github.com/asimjalis/cutest#getting-started

### DIFF
--- a/cextern/.gitignore
+++ b/cextern/.gitignore
@@ -1,0 +1,2 @@
+astrometry.net/test_healpix
+astrometry.net/example
--- a/cextern/README.md
+++ b/cextern/README.md
@@ -1,0 +1,13 @@
+# astropy-healpix/cextern/
+
+The `astropy-healpix` Python package is a wrapper around a C library.
+See http://astropy-healpix.readthedocs.io/en/latest/about.html
+
+This README gives some technical details on the C code here.
+
+- The main file is `healpix.h` and `healpix.c`, start reading there first.
+- For the Python `astropy-healpix` packge, the C code is built via `setup.py`
+- However, to help work on the C code and test it directly, a `Makefile`
+  is included here.
+- For testing, a copy of `CuTest.h` and `CuTest.c` from here is bundled:
+  https://github.com/asimjalis/cutest
--- a/cextern/astrometry.net/CuTest.c
+++ b/cextern/astrometry.net/CuTest.c
@@ -1,0 +1,340 @@
+#include <assert.h>
+#include <setjmp.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <math.h>
+
+#include "CuTest.h"
+
+/*-------------------------------------------------------------------------*
+ * CuStr
+ *-------------------------------------------------------------------------*/
+
+char* CuStrAlloc(int size)
+{
+	char* newStr = (char*) malloc( sizeof(char) * (size) );
+	return newStr;
+}
+
+char* CuStrCopy(const char* old)
+{
+	int len = strlen(old);
+	char* newStr = CuStrAlloc(len + 1);
+	strcpy(newStr, old);
+	return newStr;
+}
+
+/*-------------------------------------------------------------------------*
+ * CuString
+ *-------------------------------------------------------------------------*/
+
+void CuStringInit(CuString* str)
+{
+	str->length = 0;
+	str->size = STRING_MAX;
+	str->buffer = (char*) malloc(sizeof(char) * str->size);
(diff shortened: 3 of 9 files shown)


dstndstn:
in the astrometry.net repository we just ship cutest.[ch];
https://github.com/dstndstn/astrometry.net/blob/master/util/cutest.c

cdeil:
OK, I've bundled cutest and added a minimal README here.

`make` still doesn't run, now I get a linker error:
https://gist.github.com/cdeil/17737eb5615acb04476a9aae42faf9aa
Maybe just a function name change?

I'm off now, but I'll continue here tomorrow morning.

cdeil:
I've fixed all the linking errors in eadbafbcdb5a7ec3dae13bee90bee2f7e6d07059 .

@astrofrog - They were mostly the result of changing functions to include the "l" in the name.
Note that some of the functions listed in  `healpix.h` were not defined in `healpix.c`, I've removed those.

I don't understand why we sometimes have functions with "l" using `int64_t`, sometimes without "l" using `int`, and sometimes both:
```
int xyzarrtohealpix(const double* xyz, int Nside);

int64_t xyzarrtohealpixl(const double* xyz, int Nside);
```

@astrofrog - Is this a remnant of testing the difference?
OK if I remove the duplicates where we have two?
Should we use `int` or `int64_t` consistently?
(of course we would only merge if all existing tests pass)

The diff here is already pretty large, so @astrofrog or @dstndstn - if these changes look OK to you, maybe we can merge, and I'll start with a new PR?

dstndstn:
Really, we should use type uint64_t.  That allows a maximum nside of
1,239,850,262, which means the healpixels are about 170 micro-arcseconds on
a side.

cdeil:
I merged this in now, so that we can continue with small follow-up PRs with smaller diff from `master`.

> Really, we should use type uint64_t.

Agreed, let's use that. Is it supported by all C compilers?
(Although I think in practice very few people will run into limits with `int64_t` but would be OK with `uint64_t`.

lpsinger:
> Really, we should use type uint64_t.  That allows a maximum nside of
> 1,239,850,262, which means the healpixels are about 170 micro-arcseconds on
> a side.

This would comes with a significant downside: you would not be able to use the returned values for indexing Numpy arrays, because the Numpy index type `np.intp` is signed.

cdeil:
Then how about defining a type alias to `int64_t` in a single place, with a comment above it that we use it because of Numpy? (or in any case document somehow why we use `int64_t` and that it shouldn't be changed.

lpsinger:
If we had a generic or templated implementation, the best possible course would be to provide routines for all integer types, and make all of them known to Numpy.

cdeil:
> If we had a generic or templated implementation, the best possible course would be to provide routines for all integer types, and make all of them known to Numpy.

Maybe continue that question over in #72 ?

Cython [fused types](http://docs.cython.org/en/latest/src/userguide/fusedtypes.html) or [cast=True](http://kylebarbary.com/blog/cython-and-multiple-numpy-dtypes/) could be useful for us, maybe a reason to use Cython and not just plain C?

Basically, I'm hoping for you @lpsinger or @astrofrog to make a recommendation for the Python interface how it should be done exactly.

lpsinger:
> Cython fused types or cast=True could be useful for us, maybe a reason to use Cython and not just plain C?

If we are using Numpy ufuncs (whether written in C or Cython), then we would just register loop functions for all of the supported types.

> Basically, I'm hoping for you @lpsinger or @astrofrog to make a recommendation for the Python interface how it should be done exactly.

Either C++ templates or C preprocessor macros would ease the implementation.

cdeil:
> Either C++ templates or C preprocessor macros would ease the implementation.

@astrofrog - As far as I know, in Astropy core only C is used (e.g. WCSLib or ERFA) and for ERFA Jinja templating of C code. Not sure if we want to do Jinja C templating, but using C++ might be a no-go, given that `astropy.healpix` only is a small package, and really even casting to one type when calling into C is probably acceptable performance.

PR #93, opened 2018-06-26
Use 'long double' for x,y,z arguments

dstndstn:
- use cosl,sinl for radec_to_healpixlf
- compute 1-vz in long double

### DIFF
--- a/cextern/astrometry.net/healpix.c
+++ b/cextern/astrometry.net/healpix.c
@@ -714,7 +714,7 @@ void healpixl_get_neighbours(int64_t pix, int64_t* neighbour, int Nside) {
         }
 }
 
-static hp_t xyztohp(double vx, double vy, double vz, int Nside,
+static hp_t xyztohp(long double vx, long double vy, long double vz, int Nside,
                     double* p_dx, double* p_dy) {
     double phi;
     double twothirds = 2.0 / 3.0;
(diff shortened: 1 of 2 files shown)


cdeil:
There's a bit of a concern with this, in the sense that it probably won't fix the issue for all users: some will end up with only "double" precision still (see https://en.wikipedia.org/wiki/Long_double).

Even worse would be if there's a distribution problem where binaries (e.g. Windows from conda-forge) don't work at all for some users (e.g. some Windows machines); I don't know if that's a real concern.

I would suggest we merge tomorrow, and continue with more edits and checks in the next days.

Unless there is a more stable way to do that computation using only `double`.
@mreinicke - do you know if there is a trick to get this numerically stable?

dstndstn:
In this case it's the sin(Dec) that drops precision; Dec->90 degrees so sin(Dec)->1

mreineck:
Yes, that's why I think it should be safer to use cos(Dec) (i.e. `sqrt(x*x + y*y)`) instead of sin(Dec) (i.e. z) in this situation.

I see similar code in other places in the package... the `z2dec` function in starutil.inc for example will fail badly near the poles. I'd recommend something like

```
double xyz2dec(double x, double y, double z)
  {
  return M_PI/2 - atan2(sqrt(x*x+y*y), z);
  }
```
It's not much slower, and it works extremely reliably.

dstndstn:
I tried this in #94 - thanks for the tip! - seems to work in our one polar test case

cdeil:
@mreineck @dstndstn - Thank you!
Merging #94 now and closing this one without merging.

PR #94, opened 2018-06-26
use more stable cos(Dec) term

dstndstn:
as suggested by @mreineck in https://github.com/astropy/astropy-healpix/pull/93#issuecomment-400416163


cdeil:
@dstndstn - Thank you!

Issue #97, opened 2018-06-28
Remove use of OpenMP?


cdeil:
We introduced the use of OpenMP in astropy-healpix, to take advantage of multiple cores (see [here](http://astropy-healpix.readthedocs.io/en/latest/performance.html)). This is done using the Cython parallel and prange feature in [astropy_healpix/core_cython.pyx](https://github.com/astropy/astropy-healpix/blob/master/astropy_healpix/core_cython.pyx). Indeed, for users that care about HEALPix speed and where it works out of the box and they want to use multi-core for that process, getting the extra speed by default is great.

However, using OpenMP has a cost in terms of installation issues. We already have a user where `pip install` fails due to OpenMP issues and we weren't able to (or had time to) help him: #78
OpenMP is also listed as the first install issue here:
https://github.com/healpy/healpy/blob/master/README.rst#known-issues

I think we should remove the OpenMP usage. My arguments:
- simple install for `astropy.healpix` is extremely important, as far as I know OpenMP is not used in Astropy core yet (true?)
- there are users that don't want the now default behaviour of using all cores. E.g. many astronomers run their Python script on institute servers with 10 or 100 users, and using all cores by accident causes problems for others, or for them if their sysadmin comes to their office.
- the few users that really need the extra speed might use joblib or dask or chunk their Numpy arrays manually, i.e. use common techniques that are used e.g. also by scikit-image or scikit-learn.
- if several users complain about speed problems with `astropy.healpix`, we can always add OpenMP or some other solution later; but I don't think there are many people who need this overall.

@astrofrog @lpsinger and also @mreineck - Thoughts?

mreineck:
How difficult is it to make OpenMP optional? healpy does that, right?
(Unfortunately I'm not at all familiar with `setuptools`)

cdeil:
I guess if OpenMP is optional and off by default when someome just does `pip install astropy`, then the concern isn't as strong. But if there optionally is OpenMP support, and some people start to use it or packagers to use it when making binaries, then I think there's still a good chance that we'll get issue reports and have to spend time to support that OpenMP case.

@astrofrog - Has this discussion or even a decision concerning OpenMP already taken place for the Astropy core package, and I'm just not aware?

lpsinger:
I agree. I think that Healpy does not use OpenMP for ufunc loops; it only uses it for the spherical harmonic transforms. @zonca could confirm.

astrofrog:
[...]

PR #102, opened 2018-07-02
Disable OpenMP by default

astrofrog:
This implements one of my suggestions in https://github.com/astropy/astropy-healpix/issues/97

Basically this makes OpenMP opt-in. We can still decide to opt in by default for conda packages if the OpenMP environment there is more predictable.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,9 @@
 
 - Add ``pixel_resolution_to_nside`` function. [#31]
 
+- Disable OpenMP by default. This can be enabled during building by setting
+  ``ASTROPY_HEALPIX_USE_OPENMP=1``. [#102]
+
 0.2 (2017-10-15)
 ================
 
--- a/astropy_healpix/setup_package.py
+++ b/astropy_healpix/setup_package.py
@@ -37,6 +37,7 @@ def get_extensions():
         language="c",
         extra_compile_args=['-O2'])
 
-    add_openmp_flags_if_available(extension)
+    if os.environ.get('ASTROPY_HEALPIX_USE_OPENMP') == '1':
+        add_openmp_flags_if_available(extension)
 
     return [extension]
--- a/docs/performance.rst
+++ b/docs/performance.rst
@@ -38,19 +38,49 @@ not matter. For larger arrays, the difference is a factor of a few at most
 (without multi-threading). We will add more benchmarks over time to provide a
 more complete picture.
 
+Using OpenMP
+------------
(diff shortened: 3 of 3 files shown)


cdeil:
I would prefer if we remove OpenMP completely. This does not really resolve the concerns I mentioned in #97.

If we offer the option, we'll have packagers and users using OpenMP and have to support it, and it's a portability concern for Astropy core.

Also we have to test it. Note that this PR leaves the option to use OpenMP, but as far as I can see it's completely untested in CI now.

Also: it hasn't really been shown that this multi-threading gives good speed-ups. There have been some benchmarks posted in the PRs adding OpenMP that show a factor 2 speed-up for large arrays (and slow-down for small arrays), but not that our HEALPix C lib is not efficient yet, it can be improved by a factor of a few (see e.g. https://github.com/astropy/astropy-healpix/issues/34#issuecomment-401709090 and https://github.com/astropy/astropy-healpix/issues/34#issuecomment-401791023 ) and once it's optimised it's not clear if using many cores helps a lot.

So @astrofrog - If you want to keep the OpenMP, then I agree this is an improvement and we should merge.

astrofrog:
Removing it is fine by me as long as you are happy with the performance. I only added it originally because there were concerns that without, things were slower than healpy.

cdeil:
> Removing it is fine by me as long as you are happy with the performance. I only added it originally because there were concerns that without, things were slower than healpy.

+10 from me to remove OpenMP.

@lpsinger - Acceptable for you?

@astrofrog - Can you update this PR, or should prepare a new one to do it?

lpsinger:
> @lpsinger - Acceptable for you?

:+1:

cdeil:
Ok, then I'm closing this PR. I did open #108 which removes OpenMP.

jamienoss:
@cdeil Check out https://github.com/astropy/astropy-helpers/pull/410. You might want to use the following pattern in your root setup.py:

```python
...
# The current default is to NOT build with OpenMP support, i.e. ``disable_openmp = True``
disable_openmp = os.environ.get('ASTROPY_HEALPIX_USE_OPENMP', '0') == '0'
generate_openmp_enabled_py(NAME, disable_openmp=disable_openmp)
...
```

You won't then need the conditional around the ``add_openmp_flags_if_available()``. Additionally, once built & installed you can call ``package.openmp_enabled.is_openmp_enabled()`` to query whether the package was built with OpenMP, this can be asserted in feedstock tests etc.

Issue #91, opened 2018-06-26
Add support for uniq pixel ID scheme


cdeil:
This HEALPix "unique identifier scheme" is starting to be used widely:
- see section 3.2 in http://healpix.sourceforge.net/pdf/intro.pdf
- see section 2.3.1 in http://ivoa.net/documents/MOC/

It's used in MOC, but also in Gammapy we want to use it for multi-resolution maps, and I see @lpsinger is already using it in LIGO (see [slides](https://dcc.ligo.org/public/0149/G1800186/003/healpix-multiresolution.pdf)), so IMO we should support it in `astropy.healpix`.

We can discuss about deeper integration, but a first step is to add converter functions.
@tboch already added a Numpy implementation in https://github.com/cds-astro/mocpy/blob/5bbdf1093b25a597e996d5087f8046354e26edf7/mocpy/utils.py#L4

```python
def uniq2orderipix(uniq):
    order = ((np.log2(uniq//4)) // 2)
    order = order.astype(int)
    ipix = uniq - 4 * (4**order)
    return order, ipix

def orderipix2uniq(n_order, n_pix):
    return n_pix + ((4**n_order) << 2)
```

@tboch or @bmatthieu3 - Do you have time to add a copy here in `astropy-healpix`? If no, I could make a PR later this week as well.

Eventually this should be re-implemented in C for maximum speed / memory usage (avoiding temp array copies), but I think for now adding this implementation here with a docstring and test would good.

cdeil:
If we code those functions with `order` as input (or in any case) IMO we should also add this helper:
```python
def nside2order(nside):
    return np.log2(nside).astype(int)
```
It works up to `nside=63`.

astrofrog:
@cdeil - I agree, though I think we should include a ``round`` kwarg as for ``pixel_resolution_to_nside``

cdeil:
After thinking about this a bit more, I'd suggest to add those as standalone helper functions for now.

Deeper integration (i.e. taking `uniq` instead of the seprate `(nside, ipix)` in the existing API can be proposed / discussed, but that's a big maybe; whereas adding support via these helper functions is an immediate and clear improvement.

cdeil:
Added by @bmatthieu3 in #105 .

Issue #95, opened 2018-06-26
Rename "order" to "scheme"


cdeil:
Currently we have `order : { ‘nested’ | ‘ring’ }`.

In HEALPix, "order" is also often used for `k = log2(nside)`, so IMO this is confusing and should be changed.

I propose to replace by "scheme".

Using "ordering" which is used in FITS serialisation might be an alternative, but IMO "scheme" is better, leaving no room for confusion.

I'm happy to make a PR to adapt the few affiliated packages using this already, but given that this is alpha software, I'd suggest to just do this change and ship it with v0.3 without deprecation / supporting both for a period of time.

@astrofrog and all - thoughts?

astrofrog:
Sounds good to me!

PR #105, opened 2018-09-06
Uniq to levelipix methods

bmatthieu3:
Following astropy/astropy-healpix#91 and astropy/regions#219 I moved the utils functions from MOCPy to here. These concern the translation of uniq number referring to a specific HEALPix cell to its order/pixel index.

I use the name `level` as a replacement for order because for the moment order refers to the HEALPix scheme used (ring, nested, XY).

So the methods added are `levelipix_to_uniq`, `uniq_to_levelipix` and I also added the methods `nside_to_level` and `level_to_nside` (the last one was already here but not accessible from the user API. I think it is useful to add it to the API because for example, with MOCPy we are more thinking in terms of level/order than nside and I usually do a nside=(1 << order) operation when instanciating a new HEALPix object in MOCPy. Having a function level_to_nside would be better I think for the visibility, what do you think ?).

Thanks for the review.

### DIFF
--- a/astropy_healpix/core.py
+++ b/astropy_healpix/core.py
@@ -115,7 +115,7 @@ def nside_to_level(nside):
     nside = np.asarray(nside, dtype=np.int64)
 
     _validate_nside(nside)
-    return np.log2(nside)
+    return np.log2(nside).astype(np.int64)
 
 
 def uniq_to_levelipix(uniq):
--- a/astropy_healpix/core.py
+++ b/astropy_healpix/core.py
@@ -138,7 +138,7 @@ def uniq_to_levelipix(uniq):
     """
     uniq = np.asarray(uniq, dtype=np.int64)
 
-    level = ((np.log2(uniq//4)) // 2)
+    level = (np.log2(uniq//4)) // 2
     level = level.astype(np.int64)
     _validate_level(level)
 
--- a/astropy_healpix/core.py
+++ b/astropy_healpix/core.py
@@ -18,6 +18,8 @@ __all__ = [
     'pixel_resolution_to_nside',
     'nside_to_npix',
     'npix_to_nside',
+    'level_to_nside',
+    'nside_to_level',
     'lonlat_to_healpix',
     'healpix_to_lonlat',
     'bilinear_interpolation_weights',
@@ -77,8 +79,33 @@ def level_to_nside(level):
     nside : int
         The number of pixels on the side of one of the 12 'top-level' HEALPix tiles.
     """
+    level = np.asarray(level, dtype=np.int64)
+
     _validate_level(level)
     return 2 ** level
+
+
+def nside_to_level(nside):
+    """
+    Find the HEALPix level for a given nside. (this is given by log2(nside))
+
+    This function is the inverse of `level_to_nside`.
+
+    Parameters
+    ----------
+    nside : int
+        The number of pixels on the side of one of the 12 'top-level' HEALPix tiles.
+        Must be a power of two.
+
+    Returns
+    -------
+    level : int
+        The level of the HEALPix cells
+    """
+    nside = np.asarray(nside, dtype=np.int64)
+
+    _validate_nside(nside)
+    return np.log2(nside)
 
 
 def nside_to_pixel_area(nside):
--- a/astropy_healpix/core.py
+++ b/astropy_healpix/core.py
@@ -20,8 +20,8 @@ __all__ = [
     'npix_to_nside',
     'level_to_nside',
     'nside_to_level',
-    'levelipix_to_uniq',
-    'uniq_to_levelipix',
+    'level_ipix_to_uniq',
+    'uniq_to_level_ipix',
     'lonlat_to_healpix',
     'healpix_to_lonlat',
     'bilinear_interpolation_weights',
@@ -118,11 +118,11 @@ def nside_to_level(nside):
     return np.log2(nside).astype(np.int64)
 
 
-def uniq_to_levelipix(uniq):
+def uniq_to_level_ipix(uniq):
     """
-	Convert a HEALPix cell uniq number to its (level, ipix) equivalent
+    Convert a HEALPix cell uniq number to its (level, ipix) equivalent.
 
-    A uniq number is a 64 bit integer equals to : ipix + 4*(4**level). Please read
+    A uniq number is a 64 bits integer equaling to : ipix + 4*(4**level). Please read
     this `paper <http://ivoa.net/documents/MOC/20140602/REC-MOC-1.0-20140602.pdf>`_
     for more details about uniq numbers.
 
@@ -148,11 +148,11 @@ def uniq_to_levelipix(uniq):
     return level, ipix
 
 
-def levelipix_to_uniq(level, ipix):
+def level_ipix_to_uniq(level, ipix):
     """
-    Convert a level and HEALPix index into a uniq number representing the cell
+    Convert a level and HEALPix index into a uniq number representing the cell.
 
-    This function is the inverse of `uniq_to_levelipix`.
+    This function is the inverse of `uniq_to_level_ipix`.
 
     Parameters
     ----------
--- a/astropy_healpix/core.py
+++ b/astropy_healpix/core.py
@@ -20,6 +20,8 @@ __all__ = [
     'npix_to_nside',
     'level_to_nside',
     'nside_to_level',
+    'levelipix_to_uniq',
+    'uniq_to_levelipix',
     'lonlat_to_healpix',
     'healpix_to_lonlat',
     'bilinear_interpolation_weights',
@@ -63,6 +65,9 @@ def _validate_nside(nside):
     if not np.all(2 ** log_2_nside == nside):
         raise ValueError('nside must be a power of two')
 
+def _validate_npix(level, ipix):
+    if not np.all(ipix < (3 << 2*(level + 1))):
+        raise ValueError('ipix for a specific level must be inferior to npix')
 
 def level_to_nside(level):
     """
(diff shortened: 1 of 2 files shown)


astropy_healpix/core.py
```diff
@@ -77,10 +84,92 @@ def level_to_nside(level):
...
+
+    Returns
+    -------
+    level, ipix: int, int
+        The level and index of the HEALPix cell computed from ``uniq``.
+    """
+    uniq = np.asarray(uniq, dtype=np.int64)
+
+    level = ((np.log2(uniq//4)) // 2)
+    level = level.astype(np.int64)
```
cdeil: Why do you call `astype(np.int64)` here? You already do above, and this line should always be a no-op and can be removed, no?

astropy_healpix/core.py
```diff
@@ -77,10 +84,92 @@ def level_to_nside(level):
...
+    level : int
+        The level of the HEALPix cells
+    """
+    nside = np.asarray(nside, dtype=np.int64)
+
+    _validate_nside(nside)
+    return np.log2(nside)
+
+
+def uniq_to_levelipix(uniq):
```
cdeil: @astrofrog @lpsinger @bmatthieu3 - what about the name here?

I mean "levelipix" is not very readable, because it's non-obvious that it's two "words".
Is "uniq_to_level_ipix" better, or generally: is there a better name?

astropy_healpix/core.py
```diff
@@ -77,10 +84,92 @@ def level_to_nside(level):
...
+        The uniq number of a HEALPix cell.
+
+    Returns
+    -------
+    level, ipix: int, int
+        The level and index of the HEALPix cell computed from ``uniq``.
+    """
+    uniq = np.asarray(uniq, dtype=np.int64)
+
+    level = ((np.log2(uniq//4)) // 2)
```
cdeil: Remove the outermost `()`, it's superfluous.

astropy_healpix/tests/test_core.py
```diff
@@ -28,6 +30,35 @@ def test_level_to_nside():
...
+
+
+@pytest.mark.parametrize("level", [
+    0, 5, 10, 15, 20, 22, 25, 26, 27, 28, 29
+])
+def test_uniq_to_levelipix(level):
+    # Generate 10 HEALPix indexes for each level
+    size = 10
+    npix = 3 << 2*(level + 1)
+    ipix_arr = np.random.randint(npix, size=size, dtype=np.int64)
```
cdeil: Suggest to remove the `_arr` from the variable names here. They make the lines very long and don't help make the code more readable IMO.

astropy_healpix/core.py
```diff
@@ -77,10 +84,92 @@ def level_to_nside(level):
     nside : int
         The number of pixels on the side of one of the 12 'top-level' HEALPix tiles.
     """
+    level = np.asarray(level, dtype=np.int64)
+
     _validate_level(level)
     return 2 ** level
 
 
+def nside_to_level(nside):
```
cdeil: Should we expose this as part of the public API?

I guess "yes", even if it's just one line plus a type cast and input validation, because it leads to more readable code than calling `np.log2` all over the place directly?

But I'm not sure: @astrofrog and @lpsinger - please comment.

astropy_healpix/core.py
```diff
@@ -77,10 +84,92 @@ def level_to_nside(level):
...
+    """
+    Find the HEALPix level for a given nside. (this is given by log2(nside))
```
cdeil: The first summary line of docstrings should be short and sweet.
Remove the `(...)` part or move it to a later line, like:
```
Find the HEALPix level for a given nside.

This is given by ``level = log2(nside)``.
```

astropy_healpix/core.py
lpsinger: At least on x86, we can do better in assembly using the bsr instruction. See [my implementation here](https://git.ligo.org/lscsoft/ligo.skymap/blob/master/src/bayestar_moc.c#L36-53).

lpsinger:
Is 'level' the correct term here? I have also seen the base 2 logarithm of nside referred to as 'order', whereas what astropy_healpix calls 'order' I have seen called the 'scheme'.

cdeil:
I think the plan is to use "scheme" and "level" in astropy-healpix, and to avoid "order" completely because it's the most ambiguous?

We discussed this before (#95) but then I didn't have any time to hack on `astropy-healpix`, so it's not applied yet.

lpsinger:
> I think the plan is to use "scheme" and "level" in astropy-healpix, and to avoid "order" completely because it's the most ambiguous?

That sounds reasonable to me. Though that will definitely require a major version bump.

astropy_healpix/core.py
lpsinger: Yes please!

astropy_healpix/core.py
cdeil: @lpsinger - Is the bottleneck the `np.log2` or the whole expression?

Are you really limited by the speed of this expression in your analyses?
Or can we keep `astropy.healpix` dead simple and avoid code that might not run everywhere?

lpsinger: I haven't benchmarked it in a while. Let me check. My code also includes a fallback for non-x86 platforms that uses integer arithmetic, though I don't know if that is faster than `np.log2`.

astropy_healpix/tests/test_core.py
cdeil: @bmatthieu3 - Can you get rid of random numbers here completely? If yes, please do.
If no, please use a seed and avoid changing the global `np.random` stream.
I think we already have examples how to do that in other parts of the tests of astropy-healpix, but if not and it's not clear what I mean, let me know and I'll point out an example.

astropy_healpix/core.py
cdeil: I looked around a bit, there is no integer log 2 in Numpy.

Probably `np.log2` isn't fast for integers in the first place, and certainly the expression `(np.log2(uniq//4)) // 2` will not be fast because of temp Numpy array copies. So the first thing to do if speed matters here would be to move this to C with a simple implementation of `ilog2`(not using asm).

@lpsinger and @tboch - Is this speed really needed here? Or would it be OK to keep the `level = (np.log2(uniq//4)) // 2` because it has acceptable performance?

lpsinger: I am attaching a short benchmark program (note: rename to `test.c` to compile). It's interesting, log2 is actually an order of magnitude faster than the bit shift method, but the x86 bsrq method is an order of magnitude faster than it.

## gcc6 from MacPorts

```
$ gcc -O3 -std=gnu11 test.c && ./a.out
bitshift: 3.003659e-08 s
log2:     7.648268e-09 s
bsrq:     1.006322e-09 s
```

## clang-6.0 from MacPorts

```
$ clang -O3 -std=gnu11 test.c && ./a.out
bitshift: 3.008260e-08 s
log2:     7.997110e-09 s
bsrq:     2.159280e-09 s
```

[test.c.txt](https://github.com/astropy/astropy-healpix/files/2359000/test.c.txt)

astropy_healpix/core.py
mreineck: Hmm, doesn't np.log2() return floats?

astropy_healpix/core.py
mreineck: Explicit bit-twiddling is most likely not the fastest solution here. Assembler instructions or compiler intrinsics would be good, but unfortunately we are in Python and numpy doesn't provide something like Python's `bit_length()`.

Unfortunately `np.log2()` is problematic too, since it can be inaccurate: see https://github.com/numpy/numpy/issues/4787 . This particular problem may have been fixed, but adding 0.5 before taking the logarithm may add some extra safety.

astropy_healpix/core.py
bmatthieu3: Yes and that is why I recast the result to int. Maybe there is a log2_integer numpy function  ? or we can use bitwise operation like [on this page](https://graphics.stanford.edu/~seander/bithacks.html#IntegerLogObvious) to compute it.

astropy_healpix/tests/test_core.py
bmatthieu3: @cdeil Ok I can remove this call to ``np.random``. Just to know, what do you mean by changing the global np.random stream ? and especially why do we do not want this ?
Is this for the following reason ? :
Some tests that are called after mine are waiting for specific rand results (based on the default internal seed maybe) that would change if I make a new call to ``np.random.randint`` before them ?

astropy_healpix/core.py
```diff
@@ -77,10 +84,92 @@ def level_to_nside(level):
...
+
+    Returns
+    -------
+    level : int
+        The level of the HEALPix cells
+    """
+    nside = np.asarray(nside, dtype=np.int64)
+
+    _validate_nside(nside)
+    return np.log2(nside)
```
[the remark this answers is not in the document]
cdeil: yes

astropy_healpix/core.py
cdeil: Sorry, I didn't realise this returned a float. Yes, we should make an int.
Numpy doesn't have a log2_integer function.

Either keep as-is, or move the whole function to C and (for now) expose it via a Cython wrapper, if the speed is needed here. Given that @lpsinger and @mreineck comment a lot about performance here, and that this function might be performance-critical e.g. for MOC, that might be the way to go.

Personally I'm OK to merge as-is and leave that as a TODO comment for later, but if you have time to work on this now, you might as well do that change already in this PR.

astropy_healpix/tests/test_core.py
cdeil: If you just call functions like `np.random.random()` from all over the place (i.e. separate tests), then you will get different numbers depending on test order execution or just the machine you're on.
Then you might get numbers and failing tests at arbitrary times, which is annoying.

You can call `np.random.seed(0)` each time before generating random numbers, or create a `rng = RandomState(seed)` and call the methods on that, or use `with NumpyRNGContext(seed)`. It seems in Astropy all of those patterns are used:
https://github.com/astropy/astropy/search?q=random&unscoped_q=random

Looks like in astropy-healpix we already have one "rogue" `np.random` call, that should be cleaned up as well:
https://github.com/astropy/astropy-healpix/blob/528292bc2bff01563d6c696880020345cde88d40/astropy_healpix/tests/test_healpy.py#L285

IMO the best solution is to not use random numbers at all in almost all cases. If you write an explicit list of numbers as test cases and manually choose 1 - 10 numbers well to cover what you want to test, it's just better than producing random numbers, becuase it's more explicit and you know what is being tested, e.g. which range of numbers.

bmatthieu3:
I did some changes. I do not quite understand for the moment why [this test](https://travis-ci.org/astropy/astropy-healpix/jobs/425708174) is failing because it concerns healpy.vec2ang ?!

Concerning the change of ``np.log2`` (or not) I do not have a strong opinion. From what I understand the options are :

- Keep it like this but we have seen there can be some floating point errors on Windows ? So maybe np.round the result of the np.log2 and recast it to int64 because ``np.around`` keeps returning floats. (or do python bitwise operation to compute the most significant bit of ``uniq // 4`` in Python but this could be slower too).
- Call C code for computing the MSB ? If we choose this option should we do it in a new PR ?

cdeil:
I would suggest to merge this, and if someone has time, do the change to a C implementation in a follow-up PR. But we should wait at least for @lpsinger to comment on that point, and also @mreineck and @astrofrog - let us know if you think the current implementation is not OK and we really should put a different one from the start.

The vec2ang fail is unrelated, see #107 . I'm not quite sure how to fix it, maybe there's a way to tell hypothesis to not generate a test for `(0, 0, 0)`, which doesn't make sense as a test case.

cdeil:
Just one more comment about preformance: note that we're doing these `validate_` calls also everywhere that loop on the whole input arrays once or twice, if we keep those we anyways will never get max performance.

cdeil:
I just merged #107 and will restart the build here.

cdeil:
I'll merge this now, so that @bmatthieu3 can continue with https://github.com/astropy/regions/pull/219 to put MOC in regions and use these functions from here.

Of course, new issues or PRs to change or improve these functions are welcome, this doesn't have to be the final version. Ideally we'd find a version that's fast and platform / compiler independent.
And we might want to add a `check_inputs=True` option, which power-users can use to skip the input validate functions like the `_validate_npix` introduced in this PR, if they want maximum speed.

PR #107, opened 2018-09-08
Test fails in master

cdeil:
I was looking at https://travis-ci.org/astropy/astropy-healpix/builds/425328133 for #105 and saw some fails. Actually I think they are unrelated to that PR and will show up for all PRs (cc @bmatthieu3 and @lpsinger)

E.g. `test_vec2ang` fails here:
https://travis-ci.org/astropy/astropy-healpix/jobs/425328140#L1119

And `test_uniq_to_levelipix` fails here:
https://travis-ci.org/astropy/astropy-healpix/jobs/425328145#L1220

And the style check build also fails here:
https://travis-ci.org/astropy/astropy-healpix/jobs/425328147#L860

@bmatthieu3 and @lpsinger - if you have time to fix those, please send a PR. Otherwise I can do it tomorrow evening.

### DIFF
--- a/astropy_healpix/tests/test_healpy.py
+++ b/astropy_healpix/tests/test_healpy.py
@@ -206,7 +206,11 @@ def test_boundaries_shape():
     assert b1.shape == b2.shape
 
 
-@given(vectors=arrays(float, (3,), elements=floats(-1, 1)),
+def not_at_origin(vec):
+    return np.linalg.norm(vec) > 0
+
+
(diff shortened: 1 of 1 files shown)


cdeil:
I don't think there's a use case for calling `vec2ang` with `(0, 0, 0)`. This just showed up because we use hypothesis to generate test cases and the setup for that test won't done carefully in that case. Should be fixed via  57587a0, although I never saw the fail locally, so let's see what happens in CI.

PR #113, opened 2018-12-18
Healpix rangesearch cleanup

lpsinger:
Fix casts and eliminate a memory leak.


cdeil:
Thanks!

cextern/astrometry.net/healpix-utils.c
```diff
@@ -145,7 +145,7 @@ int64_t healpix_rangesearch_radec_simple(double ra, double dec, double radius,
     ll* hps = ll_new(256);
     radecdeg2xyzarr(ra, dec, xyz);
     hp_rangesearch(xyz, radius, Nside, hps, (anbool)approx);
-    *indices = (int64_t *)malloc((int64_t)ll_size(hps) * sizeof(int64_t));
```
mreineck: Just curious: does this fix the problems you are seeing on MIPS? This change shouldn't cause different behavior under any realistic circumstances, correct?

lpsinger: Yes, it did fix the errors on MIPS.

PR #132, opened 2019-07-03
Remove old unused helper functions

cdeil:
This PR removes the old and unused `_restore_shape` and `_validate_healpix_index` helper functions. I just noticed them in the code coverage report.

There is the question if we should validate `healpix_index` or not, and if it's not in the range what to do. (the `_restore_shape` just isn't needed any more since @lpsinger introduced ufuncs)

Currently the check is here:
https://github.com/astropy/astropy-healpix/commit/87bf0c1990825f51ac4fa5a289461b50608ddcd0#diff-d16b2ef86578243bf558a029cbcd7fdeR33
and what we get is `NaN` in the output and a RunTimeWarning:
```
In [5]: astropy_healpix.healpix_to_lonlat(99, 1)                                                                                                                                                 
/Users/deil/work/code/astropy-healpix/astropy_healpix/core.py:376: RuntimeWarning: invalid value encountered in healpix_ring_to_lonlat
  lon, lat = func(healpix_index, nside, dx, dy)
Out[5]: (<Longitude nan rad>, <Latitude nan rad>)
```
Equivalent to what healpy does.

Except for the uniq ipix conversion functions, there we currently raise a ValueError when it's out of range:
https://github.com/astropy/astropy-healpix/pull/105/files#diff-97d939e5cd1cbdc8cccf59e6a607ac25R71

Probably we should be uniform, and not emit a warning and give an output in one case, and raise an error in the other case?

That's a usability question and probably the more important one. But there's also the performance concern - doing if / else for every pixel in the hot for loop is probably a large slowdown?

@lpsinger @astrofrog @bmatthieu3 - Thoughts?

### DIFF
--- a/astropy_healpix/core.py
+++ b/astropy_healpix/core.py
@@ -28,20 +28,6 @@ __all__ = [
 ]
 
 
-def _restore_shape(*args, **kwargs):
-    shape = kwargs['shape']
-    if shape:
-        if len(args) > 1:
-            return [arg.reshape(shape) for arg in args]
-        else:
-            return args[0].reshape(shape)
-    else:
-        if len(args) > 1:
-            return [arg.item() for arg in args]
-        else:
-            return args[0].item()
-
-
 def _validate_order(order):
     # We also support upper-case, to support directly the values
     # ORDERING = {'RING', 'NESTED'} in FITS headers
@@ -52,12 +38,6 @@ def _validate_order(order):
         return 'ring'
     else:
         raise ValueError("order must be 'nested' or 'ring'")
-
-
-def _validate_healpix_index(label, healpix_index, nside):
-    npix = nside_to_npix(nside)
-    if np.any((healpix_index < 0) | (healpix_index > npix - 1)):
-        raise ValueError(f'{label} must be in the range [0:{npix}]')
 
 
 def _validate_offset(label, offset):


cdeil:
I see that @bmatthieu3 in `cdshealpix` simply raises `ValueError` for invalid HEALPix indices:
```
In [6]: healpix_to_lonlat(-10, depth=12)                                                                                                                                                         
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-cdb423008bc9> in <module>
----> 1 healpix_to_lonlat(-10, depth=12)

~/software/anaconda3/envs/astropy-healpix/lib/python3.6/site-packages/cdshealpix/nested/healpix.py in healpix_to_lonlat(ipix, depth, dx, dy)
    163 
    164     ipix = np.atleast_1d(ipix)
--> 165     _check_ipixels(data=ipix, depth=depth)
    166     ipix = ipix.astype(np.uint64)
    167 

~/software/anaconda3/envs/astropy-healpix/lib/python3.6/site-packages/cdshealpix/nested/healpix.py in _check_ipixels(data, depth)
     10     npix = 12 * 4 ** (depth)
     11     if (data >= npix).any() or (data < 0).any():
---> 12         raise ValueError("The input HEALPix cells contains value out of [0, {0}]".format(npix - 1))
     13 
     14 def lonlat_to_healpix(lon, lat, depth, return_offsets=False):

ValueError: The input HEALPix cells contains value out of [0, 201326591]
```

I like that solution, suggest we adopt it for `astropy-healpix`, i.e. keep the `_validate_healpix_index` function we had, call it everywhere `ipix` is passed in, and remove the `if` and check in the C code.

I think `ValueError` is nicer than `RuntimeWarning`, and there's no good use case for having invalid `ipix` entries - people that have this should fix up their code or data files using Numpy.

@lpsinger @astrofrog  - Is that OK? Or is it important to keep the RuntimeWarning behaviour for some reason?

mreineck:
> I like that solution, suggest we adopt it for astropy-healpix, i.e. keep the _validate_healpix_index function we had, call it everywhere ipix is passed in, and remove the if and check in the C code.

I expect that "cheap" functions (something like `pix2ang`, for example) will become several times slower with this approach.
Just the line `if (data >= npix).any() or (data < 0).any():` reads `data` two times, creates and fills two temporary arrays and reads them again to determine whether any element is `True`. Compared to the check inside the C loop, this is very, very expensive.

cdeil:
@mreineck - I did this one check and there the overhead of valid range check wasn't significant compared to `pix2ang`:
```
In [1]: import numpy as np                                                                                                                                                                       

In [2]: import healpy as hp                                                                                                                                                                      

In [3]: ipix = np.arange(int(1e6))                                                                                                                                                               

In [4]: %timeit np.any((ipix < 0) | (ipix > 999))                                                                                                                                                
915 µs ± 9.99 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [5]: %timeit hp.pix2ang(1, ipix)                                                                                                                                                              
/Users/deil/software/anaconda3/envs/astropy-healpix/lib/python3.6/site-packages/healpy/pixelfunc.py:528: RuntimeWarning: divide by zero encountered in _pix2ang_ring
  theta, phi = pixlib._pix2ang_ring(nside, ipix)
/Users/deil/software/anaconda3/envs/astropy-healpix/lib/python3.6/site-packages/healpy/pixelfunc.py:528: RuntimeWarning: invalid value encountered in _pix2ang_ring
  theta, phi = pixlib._pix2ang_ring(nside, ipix)
23 ms ± 153 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
Also we can move  the check to C and optimise it (use min /max instead of if?) and only pass the array once. And the `if` in the `pix2ang` loop that does this check can be removed, speeding it up.

So is there really a performance difference / concern?

@mreineck - Do you know if there is a use case to ever have `ipix` arrays with invalid entries? Or is this  something that no-one needs and that only occurs if users make mistakes in their scripts?

mreineck:
Thanks, @cdeil, for running this!
The situation is better than I had assumed, but I'd still advise not to use the proposed check. Even if the loads/stores are not a real problem in this context, the amount of memory accesses (which are out-of cache for somewhat large arrays) put a load on the memory subsystem, and this does not scale on a multi-core machine. That is, runnig this code on multiple threads will not give much of a speedup, and even if you are running only on a single thread, it will probably hurt performance of other processes running at the same time.

> So is there really a performance difference / concern?

Reducing the array size to 1000, I measure:
```
In [42]: ipix = np.arange(int(1e3))

In [43]: %timeit np.any((ipix < 0) | (ipix > 999))  
The slowest run took 20.55 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 5: 6.21 ��s per loop

In [44]: %timeit hp.pix2ang(1, ipix)
/home/martin/.local/lib/python3.7/site-packages/healpy/pixelfunc.py:528: RuntimeWarning: divide by zero encountered in _pix2ang_ring
  theta, phi = pixlib._pix2ang_ring(nside, ipix)
/home/martin/.local/lib/python3.7/site-packages/healpy/pixelfunc.py:528: RuntimeWarning: invalid value encountered in _pix2ang_ring
  theta, phi = pixlib._pix2ang_ring(nside, ipix)
The slowest run took 7.14 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 5: 33.7 ��s per loop
```

So the overhead is becoming more and more significant for smaller arrays.
I can't tell whether this is a real problem though.

>@mreineck - Do you know if there is a use case to ever have ipix  arrays with invalid entries? Or is this  something that no-one needs  and that only occurs if users make mistakes in their scripts?

I haven't ever seen such a use case, and Healpix C++ does not have this kind of check at all.

mreineck:
Just tested the size-1000 case with `pyHealpix` instead of `healpy`; there the cost of the check is already 30% of the `pix2ang` runtime.

cdeil:
> I haven't ever seen such a use case, and Healpix C++ does not have this kind of check at all.

Then one thing I'd like to try is to remove all the checks here and see what happens:

https://github.com/astropy/astropy-healpix/blob/a659447630015bf25d248841011d3d6a699f7200/astropy_healpix/_core.c#L65-L76

That should improve the speed quite a bit.

And then ipix range validation could be done via a single array loop ufunc without array copies that is called as an extra function before, and if really anyone has performance concerns for this, we could add a flag `check_range` with default `True` that the few users that need absolute max speed can change to `check_range=False`.

I don't have time this week, and am at meetings the next two weeks, so there's plenty of time for anyone to chime in here, but then we should converge and decide on a uniform pattern for range checks and implement it.

astrofrog:
Since @cdeil doesn't have time to work on this anymore and the original changes in this PR still make sense, I've just rebased this and will merge if it passes.

lpsinger:
Oh yeah. I forgot to weigh in here. I *think* that the new ufuncs do the pixel valid checks for you, so no need for the checks in C.

lpsinger:
> so no need for the checks in C.

Ahem. I mean, no need for the checks in Python.

lpsinger (approved):

PR #141, opened 2020-05-10
Add core wrappers for healpix_to_xyz and xyz_to_healpix

lpsinger:
The core Astrometry.net implementations of lonlat_to_healpix and healpix_to_lonlat do conversion between spherical coordinates and Cartesian coordinates. If the user wants Cartesian coordinates, then we can save some trigonemetric function evaluations by providing functions that natively speak Cartesian coordinates.

### DIFF
--- a/astropy_healpix/_core.c
+++ b/astropy_healpix/_core.c
@@ -106,6 +106,72 @@ static void lonlat_to_healpix_loop(
 }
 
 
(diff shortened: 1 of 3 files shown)


lpsinger:
I find myself in a position where it would be handy to have this function available. Would you be able to do a release soon?

lpsinger:
Asking again for a release...

PR #142, opened 2020-05-15
Update package infrastructure to follow APE 17 guidelines

astrofrog:
This also switches the CI to use Azure Pipelines, since it's very easy to configure using the OpenAstronomy pipelines. I've also included the auto-release pipeline which means that releases can be triggered by simply pushing a tag, as described in ``RELEASING.rst``. I still need to set up the adequate permissions on Azure so that publishing the releases actually works (edit: done!)

I'll disable CircleCI if/when this is merged.


astrofrog:
@lpsinger - just to check, any objections to this?

lpsinger:
I glanced quickly through the patch and it looks fine. I haven't updated my own package to APE 17 yet, so I am not an expert in the new package layout yet.

astrofrog:
@lpsinger - no problem, thanks for taking a quick look!

PR #147, opened 2020-10-05
Add HEALPix.level property

lpsinger:


astrofrog (changes requested):
Looks good! Could you add an entry in the changelog? Thanks!

lpsinger:
> Looks good! Could you add an entry in the changelog? Thanks!

Done.

PR #166, opened 2021-11-26
Disable Azure pipelines for Python 3.6

lpsinger:
Python 3.6 is apparently no longer supported. That's fine, because 3.6 hits end of life in a few weeks.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@ Changes
 - The ``frame`` keyword argument for the high-level ``HEALPix`` class may now
   be a frame name, frame instance, or frame class. [#156]
 - On instantiation, the ``HEALPix`` class checks the ``order`` argument. [#162]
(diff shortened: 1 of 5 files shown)


astrofrog:
Thanks! We should also bump the minimum Python in setup.cfg I think, if we aren't going to be testing 3.6.

We should also start building wheels for 3.10 but that could be done in a separate PR?

lpsinger:
> Thanks! We should also bump the minimum Python in setup.cfg I think, if we aren't going to be testing 3.6.

Good point. Done. That will probably fix the failing cibuildwheel job.

> We should also start building wheels for 3.10 but that could be done in a separate PR?

Yes.

astrofrog:
I've pushed a commit to try and skip the musllinux builds

lpsinger:
FYI, this is passing now...

pllim:
Would be nice if someone merges this...

lpsinger:
Self-merging. 👀

PR #168, opened 2021-12-31
MNT: Stop using deprecated astropy.tests functions

pllim:
This pull request is to remove soon to be deprecated code; also see astropy/astropy#12633 . Also stop using `distutils` and officially set minversion for `astropy`.

### DIFF
--- a/astropy_healpix/conftest.py
+++ b/astropy_healpix/conftest.py
@@ -7,19 +7,11 @@ import os
 
 import numpy as np
 
-from astropy.version import version as astropy_version
-
-# For Astropy 3.0 and later, we can use the standalone pytest plugin
-if astropy_version < '3.0':
-    from astropy.tests.pytest_plugins import *  # noqa
-    del pytest_report_header
+try:
+    from pytest_astropy_header.display import PYTEST_HEADER_MODULES, TESTED_VERSIONS
     ASTROPY_HEADER = True
-else:
-    try:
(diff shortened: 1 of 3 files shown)


pllim:
One could also turn all unhandled warnings into exceptions for `pytest` in `setup.cfg` but that is out of scope here.

pllim:
I think failures are addressed in  #166

lpsinger:
> I think failures are addressed in #166

Merged that. Would you please rebase this PR to check that the tests pass?

pllim:
I rebased but I don't understand why Azure won't kick off.

```
This check was skipped

No pipeline branch filters matched the modified branch.
```

lpsinger:
I don't understand either. Do you recall, were all of the tests except for the Python 3.6 ones passing? If so, then I'm happy to merge.

lpsinger (approved):

pllim:
Yes, I think so. Well, the linux wheel also failed but I think that is because of Python 3.6 as well.

https://dev.azure.com/astropy-project/astropy-healpix/_build/results?buildId=3568&view=results

pllim:
p.s. Maybe should just ditch Azure and use Actions.

pllim:
Re: template -- Depends. I see that someone set up wheels stuff on your Azure pipeline. So you can start with template but that might not be enough.

https://github.com/astropy/package-template/blob/cookiecutter/%7B%7B%20cookiecutter.package_name%20%7D%7D/.github/workflows/ci_tests.yml

PR #173, opened 2022-05-23
Fix some broken remote data doctests

lpsinger:


pllim (approved):
As long as the CI is happy, LGTM. I don't maintain this package so I have no idea what these changes really mean.

PR #178, opened 2022-06-30
WIP

nstarman:
I don't have time to work on this much in the foreseeable future. So feel free to take this over the finish line or leave as is, no rush either way. 
This PR is an early and incomplete implementation of healpix as a Representation. No more need for special methods like ``xyz_to_healpix``. Using the representation machinery a healpix rep can transform to any Astropy representation, e.g. PhysicsSpherical. Also, since Representations can be inside a Frame or SkyCoord, this is super easy to work with. 
If implemented, this would be nice to use as the backend in ``HEALPix``. IMO there's a lot of functions in this package that can be organized into methods on data structures. I've used some of them here, but the actual implementation could be moved to the Representation object and the separate function deprecated.

Signed-off-by: nstarman <nstarkman@protonmail.com>

### DIFF
--- a/astropy_healpix/representation.py
+++ b/astropy_healpix/representation.py
@@ -1,0 +1,178 @@
+import astropy_healpix
+from abc import abstractmethod
+from astropy.coordinates.representation import (
+    REPRESENTATION_CLASSES,
+    DUPLICATE_REPRESENTATIONS,
+    BaseRepresentation,
+    CartesianRepresentation,
+)
+from astropy.coordinates import Distance
+import astropy.units as u
+
+
+class HEALPixRepresentationBase(BaseRepresentation):
+
+    attr_classes = {
+        "indices": u.Quantity,
+        "dx": u.Quantity,
+        "dy": u.Quantity,
+    }
+
+    _order: int
+    _nside: int
+    _dx: u.Quantity
+    _dy: u.Quantity
+
+    @classmethod
+    def get_subclass(cls, *, order, nside):
+        name: str = f"HEALPix{order.capitalize()}{nside}"
+        if "Unit" in cls.__name__:
+            name += "Unit"
+
+        if name.lower() in REPRESENTATION_CLASSES:
+            return REPRESENTATION_CLASSES[name.lower()]
+        elif "abc." + name in REPRESENTATION_CLASSES:
+            return REPRESENTATION_CLASSES["abc." + name]
+
+        return type(name + "Representation", (cls,), {}, order=order, nside=nside)
+
+    @classmethod
+    @property
+    @abstractmethod
+    def order(cls):
+        return cls._order
+
+    @classmethod
+    @property
+    @abstractmethod
+    def nside(cls):
+        return cls._nside
+
+    def __init__(self, *args, differentials=None, copy=True):
+        super().__init__(*args, differentials=differentials, copy=copy)
+        self._indices = (self._indices << u.one).astype(int)
+        self._dx <<= u.one
+        self._dy <<= u.one
+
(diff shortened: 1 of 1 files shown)


lpsinger:
I love this idea! I tried to do something like this many years ago, but never got it over the finish line. I am excited to review this.

pllim:
@lpsinger , do you want this in your 1.0 ?

lpsinger (changes requested):
- [ ] This needs documentation.
- [ ] Can we support `nside` being an array? This would be important for supporting multi-resolution data sets.
- [ ] Expose pixel area, resolution, etc. as instance methods.
- [ ] Indices should be of dtype `np.intp` so that they are suitable for indexing. What do we need to do to enforce that? Currently they are a (presumably dimensionless) `u.Quantity`.
- [ ] What should the units for `dx` and `dy` be?

lpsinger:
> @lpsinger , do you want this in your 1.0 ?

Hmm, I forgot about this! Do you think that this would eventually replace the existing high-level HEALPix class?

astropy_healpix/representation.py
```diff
@@ -0,0 +1,178 @@
...
+        "dx": u.Quantity,
+        "dy": u.Quantity,
+    }
+
+    _order: int
+    _nside: int
+    _dx: u.Quantity
+    _dy: u.Quantity
+
+    @classmethod
```
nstarman, line 26: I regret making this a classmethod. This should be a standalone function.

nstarman:
Thanks @lpsinger. I opened this PR mostly as a demonstration; I think it's a good idea, but requires a fair bit of work and time I don't have ATM. I'm happy to chat / review PRs.

> * [ ]  This needs documentation.

Agreed.

> * [ ]  Can we support `nside` being an array? This would be important for supporting multi-resolution data sets.

Yes, with caveats. The `Representation` machinery requires its subclasses to have the coordinates as the only constructor arguments. e.g. ``x,y,z -> rho, theta,phi``. In the Healpix Representation, these are `indices, dx, dy`. 
`nside` and `norder` are not coordinates, but part of the definition of the Representation type, which is why they are set by `get_subclass`. So a `HealpixRepresentation[nside=4, order=6]` is a different representation type then `HealpixRepresentation[nside=4, order=8]`, though their constructor arguments are the same.
We can absolutely relax the requirement that `nside` is a single `int` to instead ``nside: int | tuple[int, ...]``. The only complication is then that `nside` would have to be the same length as the coordinates — `len(nside) == len(indices`. So the class type `HealpixRepresentation[nside=(4, 5, 6, 8), order=6]` can only accept coordinate arrays of length 4, matching `nside`.

> * [ ]  Expose pixel area, resolution, etc. as instance methods.

Sounds good.

> * [ ]  Indices should be of dtype `np.intp` so that they are suitable for indexing. What do we need to do to enforce that? Currently they are a (presumably dimensionless) `u.Quantity`.

Aren't they already set as such? In ``__init__`` I have

```python
self._indices = (self._indices << u.one).astype(int)
```

So indices is a Quantity with dtype `int`.


> * [ ]  What should the units for `dx` and `dy` be?

Currently the Representation machinery requires all coordinates to be `Quantity`. I have the units set to ``dimensionless_unscaled`` ATM, but that can be changed. Should they be `radians`?

lpsinger:
> Thanks @lpsinger. I opened this PR mostly as a demonstration; I think it's a good idea, but requires a fair bit of work and time I don't have ATM. I'm happy to chat / review PRs.

OK, in that case, I'll proceed with releasing v1.0.0 with what we have now.

PR #185, opened 2023-07-27
Opt in to Python limited API

lpsinger:


pllim (commented):
I think @astrofrog should look at this, not me. I am not familiar with the actual algorithms in this package.

astrofrog (approved):
This looks good to me in general though I wonder if we should make it so that the publish workflow can run on pull requests optionally if a specific label is set (e.g. 'Build All Wheels' or 'Extra CI' like we have in e.g. the core package) - this is especially useful for PRs like this where we need to check that the wheel building is working fine.

lpsinger:
> This looks good to me in general though I wonder if we should make it so that the publish workflow can run on pull requests optionally if a specific label is set (e.g. 'Build All Wheels' or 'Extra CI' like we have in e.g. the core package) - this is especially useful for PRs like this where we need to check that the wheel building is working fine.

I guess we already had that: https://github.com/astropy/astropy-healpix/blob/main/.github/workflows/publish.yml#L13

I triggered a new build with wheels.

lpsinger:
> Is the discussion at [astropy/astropy#14915](https://github.com/astropy/astropy/issues/14915) relevant here?

It would mean also dropping support for Python 3.8, since Numpy 1.25 requires Python >= 3.8. I'm OK with that, are you?

astrofrog:
Sounds good to me!

lpsinger:
> > I'm OK with that, are you?
> 
> It is not my call to make as I am not familiar with your user base. Perhaps @astrofrog can advise as a co-maintainer of this package.

It's fine with me. I've added patches to do that.

astrofrog (approved):
Looks good! Should we perhaps add back the oldestdeps jobs with a test against numpy 1.19 at runtime?

tox.ini
```diff
@@ -57,6 +57,8 @@ deps =
     astropy40: astropy==4.0.*
     astropylts: astropy==5.0.*
 
+    oldestdeps: numpy==1.19.* astropy==3.*
```
astrofrog, line 60: I think you need to define each dependency on two separate oldestdeps: lines

lpsinger: Fixed.

lpsinger:
Hey @astrofrog, @pllim, would you please review one last time? I added the oldestdeps job.

PR #186, opened 2023-07-27
Drop support for Python 3.7, which is at end of life

lpsinger:


.readthedocs.yml
pllim, line 1: Actually, RTD doesn't do Python like this anymore. See https://docs.readthedocs.io/en/stable/config-file/v2.html

Also no reason to not use Python 3.11 for docs?

lpsinger:
Rebased!

pllim (approved):
LGTM. Thanks!

PR #187, opened 2023-07-27
Fix setuptools deprecation warning about license_files

lpsinger:
Resolves this deprecation warning:

```
********************************************************************************
The license_file parameter is deprecated, use license_files instead.

By 2023-Oct-30, you need to update your project and remove deprecated calls
or your builds will no longer be supported.

See https://setuptools.pypa.io/en/latest/userguide/declarative_config.html for details.
********************************************************************************
```


pllim (approved):
LGTM. Thanks!

PR #198, opened 2023-11-09
Remove old placeholder workaround

dstansby:
Since gcc 10 is no longer supported (https://gcc.gnu.org/gcc-10/), and this workaround is causing issues with packaging on conda-forge (https://github.com/conda-forge/astropy-healpix-feedstock/pull/32), remove the workaround.

I had to bump the Ubuntu version in the oldestdeps test to pick up a newer version of gcc (gcc-9 > gcc-11)

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -53,7 +53,7 @@ jobs:
             toxenv: py311-test-alldeps-predeps
 
           - name: Python 3.9 with oldest supported version of all dependencies
-            os: ubuntu-20.04
+            os: ubuntu-22.04
             python: 3.9
             toxenv: py39-test-oldestdeps
 
--- a/astropy_healpix/_core.c
+++ b/astropy_healpix/_core.c
@@ -1,23 +1,3 @@
-/* FIXME:
- * The Numpy C-API defines PyArrayDescr_Type as:
- *
- *   #define PyArrayDescr_Type (*(PyTypeObject *)PyArray_API[3])
- *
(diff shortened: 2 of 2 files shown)


lpsinger (approved):
Actually, GCC 10 would be fine; the compiler bug existed on GCC 9 and earlier.

lpsinger:
@dstansby, do you need a release of astropy-healpix including this PR for the Conda packaging?

dstansby:
I think so, but there's a second (different) issue with the macOS builds on conda-forge too, so it's probably worth sorting that out too before doing a new release.

lpsinger:
> I think so, but there's a second (different) issue with the macOS builds on conda-forge too, so it's probably worth sorting that out too before doing a new release.

Would you please point me toward that other issue?

dstansby:
See https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=820558&view=logs&jobId=9c5ef928-2cd6-52e5-dbe6-9d173a7d951b for one of the failed macOS builds

lpsinger:
> See https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=820558&view=logs&jobId=9c5ef928-2cd6-52e5-dbe6-9d173a7d951b for one of the failed macOS builds

Ah, no good turn goes unpunished... https://github.com/numpy/numpy/pull/23944 was one of my PRs.

lpsinger:
My mistake, that was numpy/numpy#15355.

PR #199, opened 2023-11-11
Make Numpy loop function args const correct

lpsinger:
See https://github.com/numpy/numpy/pull/15355.

CC @dstansby.


pllim (commented):
I am not familiar with the code here at all, so I tagged Tom R to review instead.

dstansby (approved):
I applied this as a patch over at conda-forge, and it fixed the failed builds on macOS: https://github.com/conda-forge/astropy-healpix-feedstock/pull/32/commits/6510203501aea19ec045fbc68a1676fbaa5e4855

astrofrog (approved):

PR #200, opened 2023-11-20
Build wheels for PyPy

lpsinger:


lpsinger:
This should fix all of the conda build issues revealed by @dstansby in https://github.com/conda-forge/astropy-healpix-feedstock/pull/32. @astrofrog, would you please review this so that we can do a release to fix the conda build without adding patches?

astrofrog (approved):

PR #201, opened 2023-11-20
Unpin Numpy version for building wheels

lpsinger:
Allow using any version of Numpy between 1.25 and 2. This will allow us to use the latest release of Numpy on PyPI and take advantage of prebuilt wheels for more platforms and architectures.

Use the new `NPY_TARGET_VERSION` macro to peform a backward-compatible supporting versions of Numpy as old as 1.19.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,6 @@
 
 requires = ["setuptools>=42.0.0",
             "setuptools_scm",
-            "wheel",
             "extension-helpers",
             "numpy>=1.25,<2"]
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ requires = ["setuptools>=42.0.0",
             "setuptools_scm",
             "wheel",
```
pllim, line 5: I also don't think you need this anymore?
```suggestion
```

pllim (commented):
I'll let @astrofrog approve since I am not a maintainer of this repo but the diff LGTM. Thanks!

PR #205, opened 2023-11-28
Build wheels for musllinux

lpsinger:


pllim:
This is no longer needed. It is built now.

https://github.com/astropy/astropy-healpix/blob/2bccd28775b204c66cb467877d41aa7dc7ee69e9/.github/workflows/publish.yml#L25

PR #208, opened 2023-12-09
Test for invalid lon, lat, and xyz values

lpsinger:
See #207.


astrofrog:
Thanks, will test this out shortly!

PR #214, opened 2024-04-05
Build against Numpy 2.0.0rc1 or later

astrofrog:
We need to release new wheels of reproject compiled against Numpy 2.0.0rc1 ahead of the final Numpy 2.0.0 release, so this PR updates the build-time dependencies and also simplifies the nightly wheels configuration.


astrofrog:
@neutrinoceros - there is a CI failure here related to the copy= argument. Could you take a quick look to see if it's a change that needs to be made here or an issue with astropy?

astrofrog:
The CI should be fixed in #215 hopefully

PR #215, opened 2024-04-05
BUG: fix compatibility with numpy 2.0.0 (copy semantics)

neutrinoceros:
Unblocks #214

This isn't the only place where `copy=False` is used in the code base, but that's the only one that causes tests to fail, looking at [example logs](https://github.com/astropy/astropy-healpix/actions/runs/8568206064/job/23481585092?pr=214), so the fix is minimal. If other call sites with `copy=False` raise errors with numpy 2.0 in the future they could be adjusted similarly.

### DIFF
--- a/astropy_healpix/core.py
+++ b/astropy_healpix/core.py
@@ -28,6 +28,8 @@ __all__ = [
     'neighbours',
 ]
 
+_NUMPY_COPY_IF_NEEDED = False if np.__version__.startswith("1.") else None
+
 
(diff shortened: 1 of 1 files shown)


astrofrog (approved):
Thank you! Will merge once CI passes.

astrofrog:
Can you rebase this to see if green now?

neutrinoceros:
done 🤞🏻

PR #233, opened 2025-01-29
Add sphinx.configuration variable to fix Sphinx build

lpsinger:
See https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -14,3 +14,6 @@ python:
         - all
 
 formats: []
+
+sphinx:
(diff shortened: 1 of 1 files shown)


pllim (approved):
LGTM!

Though if you are interested, here is more but you can add it as follow-up PR if you want:

```yaml
sphinx:
  builder: html
  configuration: docs/conf.py
  fail_on_warning: true
```

PR #239, opened 2025-01-31
Modernize supported Python and Numpy versions

lpsinger:


tox.ini
```diff
@@ -49,15 +48,14 @@ deps =
...
 
     astropy50: astropy==5.0.*
     astropy60: astropy==6.0.*
 
-    oldestdeps: numpy==1.19.*
-    oldestdeps: astropy==4.0.*
+    oldestdeps: numpy==1.25.*
+    oldestdeps: astropy==5.0.*
 
     devdeps: git+https://github.com/astropy/astropy.git#egg=astropy
```
pllim, line 60: You can use nightly wheel here. Example: https://github.com/spacetelescope/synphot_refactor/blob/master/tox.ini

.github/workflows/ci_tests.yml
```diff
@@ -27,40 +27,40 @@ jobs:
...
+            python: '3.10'
+            toxenv: py310-test-oldestdeps
 
-          # - name: Python 3.10 with latest dev versions of key dependencies
+          # - name: Python 3.11 with latest dev versions of key dependencies
           #   os: ubuntu-latest
-          #   python: '3.10'
-          #   toxenv: py310-test-devdeps
+          #   python: '3.11'
+          #   toxenv: py311-test-devdeps
```
pllim, line 63: Why is this commented? Also you don't want py313 running in the matrix?

tox.ini
```diff
@@ -1,8 +1,8 @@
 [tox]
 envlist =
-    py{39,310,311}-test{,-alldeps,-devdeps-predeps}{,-cov}
-    py{39,310,311}-test-numpy{116,117,118}
-    py{39,310,311}-test-astropy{50,60}
+    py{310,311,312,313}-test{,-alldeps,-devdeps-predeps}{,-cov}
+    py{310,311,312,313}-test-numpy{125,200}
+    py{310,311,312,313}-test-astropy{50,60}
```
pllim, line 5: Are these combos actually used?

.github/workflows/ci_tests.yml
lpsinger: I think that we copied this from a package template many years ago, and these were commented out. I uncommented them and switched them to Python 3.13.

tox.ini
lpsinger: No, they are covered by oldestdeps. Removed.

tox.ini
lpsinger: Done.

lpsinger:
@pllim, all done, please take another look.

pllim (approved):
LGTM. Thanks!

PR #246, opened 2025-02-03
Update changelog for version 1.0.4

lpsinger:


pllim (approved):
Thanks!

Since you bumped minversions, are you sure this is strictly bugfix release and not a v1.1.0? Anyways, I will leave that up to you.

PR #251, opened 2025-02-19
Enable C assertions in unit tests

lpsinger:


pllim:
Please rebase to pick up fix. Thanks!

pllim:
Hmm labeled event did not kick off wheels

lpsinger:
I'll try closing and reopening it.

pllim:
For future reference, you can try to add `labeled` event to the workflow. Though it has unintended consequences if you apply many labels, so I will leave that decision up to you. Thanks!

https://github.com/astropy/astropy/blob/main/.github/workflows/publish.yml

PR #252, opened 2025-02-19
Check interpolate_bilinear_lonlat inputs for invalid values

lpsinger:
This is found to fix abort errors that occurred on macOS x86_64 when running the unit tests for the Conda packages. See https://github.com/conda-forge/astropy-healpix-feedstock/pull/41.

This might indicate that some of the Conda builds have assertions disabled via the `NDEBUG` preprocessor macro, and some have them enabled, inconsistently.

Without this patch, any number of assertions should have been tripped. For example, this one:
https://github.com/astropy/astropy-healpix/blob/v1.1.1/cextern/astrometry.net/healpix.c#L742

### DIFF
--- a/astropy_healpix/_core.c
+++ b/astropy_healpix/_core.c
@@ -258,11 +258,20 @@ static void bilinear_interpolation_weights_loop(
         double weights[4];
         int j;
 
-        interpolate_weights(lon, lat, indices, weights, nside);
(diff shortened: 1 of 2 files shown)


lpsinger:
@pllim, I have confirmed locally that merging this will also fix the build for #251.

pllim (approved):
I trust you on this one. Thanks!

PR #255, opened 2025-02-19
BLD: More explicit architectures for Linux x86

pllim:
Not sure if this helps or not but better be explicit than implicit?


pllim:
Hey hey, looks like musllinux is green. Why was it skipped before?

lpsinger (approved):

PR #283, opened 2025-12-04
Fix free-threaded Python builds

lpsinger:
The free-threaded builds were failing with the following error message:

> ValueError: `py_limited_api='cp310'` not supported. `Py_LIMITED_API` is currently incompatible with `Py_GIL_DISABLED`. See https://github.com/python/cpython/issues/111506.

### DIFF
--- a/astropy_healpix/_core.c
+++ b/astropy_healpix/_core.c
@@ -13,28 +13,6 @@
  */
 #include "ieee754.h"
 
-/* FIXME:
- * The Numpy C-API defines PyArrayDescr_Type as:
- *
- *   #define PyArrayDescr_Type (*(PyTypeObject *)PyArray_API[3])
- *
- * and then in some places we need to take its address, &PyArrayDescr_Type.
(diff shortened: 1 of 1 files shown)


pllim:
`astropy` itself does not do this, so I don't see a good reason to do it here. How can you be sure the code is truly thread safe here? I will defer to @astrofrog .

https://github.com/astropy/astropy/blob/8a4579a66c33d8ad25c3364392a0d671b1bb4784/pyproject.toml#L290

ref: https://github.com/astropy/astropy/issues/16916

lpsinger:
astropy-healpix itself is less likely than astropy to have nogil bugs because astropy-healpix's extension modules only define Numpy ufuncs that have no static or global state at all --- they are all pure functions.

My take is that we are likely to find and fix nogil bugs more quickly if we enable the builds and start testing them routinely.

lpsinger:
> I don't have objection but I would give Tom R a chance to comment. Can this wait a few days?

Of course, no hurry.

PR #284, opened 2025-12-04
Remove cibuildwheel PyPy skips

lpsinger:
PyPy wheels are not longer built by default.


pllim:
You might have to exclude "cp314t"?

PR #311, opened 2026-06-24
Fix off-by-one in uniq_to_level_ipix at level boundaries

mcoughlin:
This PR computes the level with exact integer arithmetic preventing an issue at level boundaries.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,14 @@
 Changes
 *******
 
+1.1.4 (unreleased)
+==================
+
+- Fix an off-by-one error in ``uniq_to_level_ipix`` that returned the wrong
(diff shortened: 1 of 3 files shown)


mcoughlin:
@lpsinger

lpsinger (commented):
Maybe it is time to copy over some of the compiler intrinsics and assembly instructions that I use in ligo.skymap. There is also a lot of useful functionality in stdbit.h that was introduced in C23.

lpsinger:
This will take me several days to parse through since I have some C ufunc code that I have been saving for this which I would like to compare.

lpsinger:
See #316.

pllim:
Superseded by #316

PR #315, opened 2026-06-26
Require Numpy >= 2

lpsinger:


pllim (approved):
I think this is reasonable by now. Thanks!

PR #316, opened 2026-06-26
Fix incorrect results from uniq_to_level_ipix near npix bounds

lpsinger:
Rewrite using `uniq_to_level_ipix` and `nside_to_level` using bitwise and integer arithmetic to fix all floating-point and rounding errors.

This is a reworking of #311.

### DIFF
--- a/astropy_healpix/_core.c
+++ b/astropy_healpix/_core.c
@@ -6,6 +6,12 @@
 #include "healpix-utils.h"
 #include "interpolation.h"
 
+#if __has_include(<stdbit.h>)
+    #include <stdbit.h>
+#elif defined(_MSC_VER)
+    #include <intrin.h>
+#endif
+
 /* FIXME: We need npy_set_floatstatus_invalid(), but unlike most of the Numpy
  * C API it is only available on some platforms if you explicitly link against
  * Numpy, which is not typically done for building C extensions. This bundled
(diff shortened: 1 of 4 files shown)


lpsinger:
I have tested this locally with Numpy 2 and the tests are passing, so I believe that the unit test errors in this PR will be fixed by #315. Please review #315 first.

pllim (commented):
Please rebase and also throw in a change log for https://github.com/astropy/astropy-healpix/pull/315 here. Thanks!

pllim:
Looks like it failed on Windows.

pllim:
`Windows fatal exception: access violation`

Might need to debug segfault

lpsinger:
The real fatal exception? Booting Windows 😝 

This will take some more time to debug.

lpsinger:
That last commit may fix it, changing

```c
_BitScanReverse64(index, in)
```

to

```c
_BitScanReverse64(&index, in)
```

lpsinger:
@mcoughlin, would you please take a look?

astropy_healpix/_core.c
```diff
@@ -294,6 +300,30 @@ static void neighbours_loop(
...
+        #if __has_include(<stdbit.h>)
+            *out = 63 - stdc_leading_zeros(in);
+        #elif defined(_MSC_VER)
+            unsigned long index;
+            if (_BitScanReverse64(&index, in))
+                *out = index;
+            else
+                *out = -1;
+        #else
+            *out = 63 - __builtin_clzl(in);
```
mcoughlin: *out = 63 - __builtin_clzll(in);

__builtin_clzll(in) counts how many leading zero bits are in the 64-bit value in, so 63 - that gives the position of its highest set bit.

mcoughlin (approved):
Looks good. I think you just need __builtin_clzll at one line.

astropy_healpix/_core.c
lpsinger: Do you just mean that I should replace `__builtin_clzl` with `__builtin_clzll`?

mcoughlin: I think so!

lpsinger: Right, that's guaranteed to be 64 bit on all platforms. Done.

mcoughlin (approved):
LGTM

pllim (approved):
I'll trust y'all on this one. Approving by proxy. Thanks!

PR #323, opened 2026-07-20
Make return value of nside_to_level a 64-bit int

lpsinger:
It is very common to use the return value as a bit-shift argument which would cause truncation if the user failed to first convert the return value to a 64-bit integer.

See https://git.ligo.org/emfollow/userguide/-/merge_requests/332.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,13 @@
 Changes
 *******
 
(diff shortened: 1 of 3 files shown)


lpsinger:
CC @mcoughlin

astrofrog (approved):

